PR #3, opened 2022-06-19
Run Open Mower with custom comms node

jkaflik:
Currently we have a baseline implementation of comms node for Open Mower's motherboard.

Since there is (and hopefully we will have it more soon) https://github.com/cloudn1ne/Mowgli which may become a drop-in replacement, my idea is to have a straightforward possibility to choose which comms node to run. 

Actually, there is no need to make it mowgli specific, since @cloudn1ne decided to use `rosserial` package.

By default `openmower`'s comms package will be run.

### DIFF
--- a/open_mower/launch/include/_comms.launch
+++ b/open_mower/launch/include/_comms.launch
@@ -5,12 +5,23 @@
 <launch>
     <!-- TODO: Add parameter for simulation -->
 
-    <node pkg="mower_comms" type="mower_comms" name="mower_comms">
-        <param name="ll_serial_port" value="/dev/$(env OM_LL_SERIAL_PORT)"/>
-        <param name="left_esc_serial_port" value="/dev/$(env OM_LEFT_SERIAL_PORT)"/>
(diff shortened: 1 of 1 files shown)


ClemensElflein:
Thank you for the PR!
That's a great idea! I really like the effort which is put into Mowgli.

I think I would configure it a little different, though:

Basically for the "low level" includes (hardware comms) I'd create a device specific folder which then gets included depending on OM_MOWER environment variable.

E.g. launch/include/YardForce500/_comms.launch, launch/include/YardForce500Mowgli/_comms.launch, ...

Since the change for Mowgli is so little, it might seem overkill but we'll be more flexible in the future with this approach. What do you think?

mkelk:
I'm using switches from the config file's OM_MOWER right now to switch to replacing mower_comms with my own parallel package using Mowgli and a few other things in the launch files (like starting rosserial). I think this is a fairly clean way of using other implementations. Once I hopefully get it actually mowing, I'll clean up and share.

PR #4, opened 2022-06-19
Container image

jkaflik:
Happy to share container image I use to get things done. It's based on an official ROS' noetic base image. This has to be considered for development purposes only as it is not optimised (no multistage build, hence size is big) and does not guarantee security.

Build will be automatically triggered with Github Actions and pushed with two amd64, arm64 targets

Container runs a `roslaunch open_mower open_mower.launch` by default, unless any other CMD is provided.
It's possible to run any ROS command or bash shell.

While working on this, it came to my attention that the setup of having ROS workspace in main Open Mower repository is not sustainable - as open mower package (in this repository) is not designed to run separately.
IMO this repository should contain whole ROS workspace. Walk-around for the build is to clone main repository during the image build.

Next steps:

- [ ] provide quick introduction in README
- [ ] how to run container on RPi
- [ ] how to run container on macOS (with X11 forwarding)

### DIFF
--- a/.github/assets/openmower_entrypoint.sh
+++ b/.github/assets/openmower_entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -e
+
+# setup ros environment
+source "/opt/ros/$ROS_DISTRO/setup.bash"
+source /opt/openmower/ROS/devel/setup.bash
+
+exec "$@"
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -1,0 +1,47 @@
+name: Build
+
+# Controls when the workflow will run
+on:
+  workflow_dispatch
+  push:
+    branches:
+      - 'main'
+      - 'dev'
+    tags:
+      - 'v*.*.*'
+  pull_request:
+    branches:
+      - 'main'
+      - 'dev'
+
+# permissions are needed if pushing to ghcr.io
+permissions:
+  packages: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      # Get the repository's code
+      - name: Checkout
+        uses: actions/checkout@v2
+      # https://github.com/docker/setup-qemu-action
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      # https://github.com/docker/setup-buildx-action
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to GHCR
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm/v7
+          push: ${{ github.event_name != 'pull_request' }}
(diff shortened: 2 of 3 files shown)


ClemensElflein:
Thank you for the PR!
So basically this builds a docker image with prebuilt binaries, so we can just run that as "release" without compiling anything on the device itself? That would be nice for a release.

I have no idea how to "package" a ROS app and just a basic understanding on how docker works. Won't running it on docker degrade performance? My question basically is: couldn't the docker image (since it has the compiler and everything) build the binaries and export them to a zip file which can be copied to the target and be executed directly? The compiled sources are basically just binaries and some shared libs anyways, right?

I wasn't sure how to set it up, but I didn't want to have a reference from the slic3r planner and the FTC controller to OpenMower, because it can be used in other contexts as well. That's why I used this repo for the mower specific stuff and put the ROS worspace in the OpenMower repo.

I think that the final releases should also live in the OpenMower repo, because people will expect it there and it has the workspace with the dependencies.

jkaflik:
@ClemensElflein 

> so we can just run that as "release" without compiling anything on the device itself? That would be nice for a release.

yes. However, in this case it's more like dev image, but can be used as a release one until we cleanup the project a little bit.

> Won't running it on docker degrade performance?

Docker or any other container orchestrator does not have any measurable* overhead since it's using Linux native mechanisms, like cgroup.

* unless you use it in VM like Docker for Mac does

> My question basically is: couldn't the docker image (since it has the compiler and everything) build the binaries and export them to a zip file which can be copied to the target and be executed directly? The compiled sources are basically just binaries and some shared libs anyways, right?

Yes. It's possible. Nowadays, that's exactly how most of the CI/CD is done. In this case, we can also aim for that. I can help with this, but first I need to learn how to distribute ROS packages. For now, I found having container image simplest way.

> but I didn't want to have a reference from the slic3r planner and the FTC controller

Yeah. I don't say we should have the external libs living in a repository. Git submodules are fine. What I mean, we should have one single repository with ROS workspace, instead of having workspace in a main repo and open mower packages in another one.


I happy to take further steps with that. For now, I recommend to have this MVP merged so I can work on some doc

ClemensElflein:
Thank you for the detailed reply! Sounds good! Let's start with docker, then. If we have performance issues, we can still figure out another way to deploy this thing.

True, I thought about just copying whatever is installed in /opt/ros and the binaries built by the docker image. But you are right, there might be other OS dependencies and Docker solves this. We can even run Raspbian then (even though I have no problems with Ubuntu, but Raspbian seems to be the popular choice anyways).

Not sure how ROS in a Docker will work with networking, because ROS sometimes is weird with non trivial network config, but we'll figure it out.

jkaflik:
> We can even run Raspbian then (even though I have no problems with Ubuntu, but Raspbian seems to be the popular choice anyways).

This is exactly my setup. Plus I am able to run it on my Mac M1

> Not sure how ROS in a Docker will work with networking, because ROS sometimes is weird with non trivial network config, but we'll figure it out.

Docker allows to run containers with host network, so ROS should be fine.

ClemensElflein:
Hmm unfortunately, the build fails. I don't have the time right now to check this and I think that you might know better why it's failing. If it's easier for you to debug this, I can give you write access to this repo

ClemensElflein:
@jkaflik just tagging, in case you don't get the notification on a closed PR

jkaflik:
@ClemensElflein thanks for a heads up. I will take a look on this later. I don't need write for now, I will try to sort it out on my fork

jkaflik:
it needs some extra debugging as it builds fine on my machine. probably there is some difference between docker & buildx. i will handle that

PR #16, opened 2022-08-29
Add boolean topics for recording control

mnh-jansson:
Adds boolean topics in addition to the joy subscription in the area recording behavior to be used with joy_teleop and ROS-Mobile. This also allows setting parameters for multiple gamepads and can be set with the OM_MOWER_GAMEPAD variable" 

Example of ROS-Mobile setup:
Screenshot_20220829-191310


ClemensElflein:
Very cool PR! Thank you for your work! Finally we're able to use other gamepads or an app on robots which don't have the external USB port.

PR #19, opened 2022-09-18
Update README.md

Erfre:
Path doesn't match project path.


Erfre:
This path didn't work. Updated it to match project. If it's supposed to not match reject this.

Issue #5, opened 2022-06-20
Create hardware dependent launch configurations


ClemensElflein:
Basically what was discussed in #3

jkaflik:
XML explained (please see section #3): http://wiki.ros.org/roslaunch/XML

mnh-jansson:
Unfortunately i don't think the current solution in development branch is flexible enough to support hardware variations such as Mowgli as it requires rosserial. Would it be a good idea to move over to a one liner _include_ instead of _params_?

ClemensElflein:
I also don't want mowgli and open mower to drift apart too far, I have fetched the important changes (again with some adaptations) from mowgli this week, but with time this will be hard due to small changes here and there.


## Current Approach
It can be used like that. We could have hardware dependent include directories and move all the serial interface mappings out of the mower_config.sh and have just some high level config there (how wide is my mower, how far should it undock etc).

For example:
open_mower.launch currently includes "_comms.launch" which connects to everything on the hardware (+ntrip, that should be moved out of there).

We could have
- launch/hardware/mainboard_vXYZ/_comms.launch
- launch/hardware/mowgli/_comms.launch

and so on, then we can include the right one using the OM_MOWER switch.

The only downside is that _all_ dependencies of all platforms are part of the dependency tree (i.e. rosserial will be installed even when not using mowgli and xESC stuff will be installed when using mowgli)


## Another Approach
So we also could do this as an alternative:
We move everything that is shared between all hardware platforms (i.e. the high level stuff) into one repo called open_mower_core or something (or move the hardware dependent stuff out of this one, probably even better) and then create a repo for each hardware platform that includes the core as a dependency (similar to what mowgli is doing right now).

So in the end the user running mowgli will get the mowgli repo which then gets the current compatible open mower core and I could have like open_mower_yardforce, open_mower_husqvarna etc.


Also pinging @cloudn1ne - what do you think?

cloudn1ne:
assuming you want to switch to BT also for OpenMower:

Im not fully there yet, but im positive that in a few more days i have the BT++ working in the same state as it is now with the slightly enhanced mower_logic (stop, resume) behaviors. From then on it should be really easy to add additional behaviors by creating new BT nodes.

The BT code is already written in a way that it is independent from the hardware as much as possible and where possible (so excluding map service calls) even independent from a "mower usercase", as there might be other uses for the robot platform ;-)

[...]

PR #23, opened 2023-03-31
Allow COMMS to be disabled when OM_NO_COMMS is set to true

cedbossneo:


ClemensElflein:
I think we should solve that as mentioned in #5. It's more flexible and custom implementations will need to add their own stuff anyways, so creating a custom launch file without the comms node is trivial.

PR #26, opened 2023-05-01
Docking behavior fixes

olliewalsh:
Wait for GPS in docking approach.

If docking approach fails, it's not safe to undock, so just retry docking. Reset retry count after docking approach.

When starting mowing, only undock if docked, otherwise just start mowing.

If docking needed, e.g low battery, do not abort the current behaviour if it's already docking/undocking.

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
@@ -173,6 +173,10 @@ std::string DockingBehavior::state_name() {
 }
 
 Behavior *DockingBehavior::execute() {
+    while(!isGPSGood){
+        ROS_WARN_STREAM("Waiting for good GPS");
+        ros::Duration(10.0).sleep();
+    }
 
     bool approachSuccess = approach_docking_point();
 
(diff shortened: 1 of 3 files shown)


ClemensElflein (changes requested):

ClemensElflein:
Thank you for the PR, the changes look good for me, I'd just propose a small change for checking for the current state (see comment above)

src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
```diff
@@ -174,21 +174,29 @@ std::string DockingBehavior::state_name() {
 
 Behavior *DockingBehavior::execute() {
 
+    if(!isGPSGood){
```
olliewalsh, line 177: should call it

olliewalsh: works without it so will remove for now

olliewalsh: ah, it gets reset when entering the instance. Need to just wait in a loop instead

olliewalsh:
seems to be working ok but set to draft until I can test this a couple of more times

ClemensElflein:
alright, looks good to me. let me know once it's confirmed working

olliewalsh:
Ran for a day with zero docking issues. Previously would fail quite often, depended on the dock approach distance as it was a race to dock before the voltage drops back down below the threshold

ClemensElflein:
sounds good, you should probably increase the empty voltage threshold though

olliewalsh:
yeah, I did increase it a little when not trying to reproduce this (there is an issue with that, will submit a PR soon)

PR #27, opened 2023-05-06
added mow motor current to the monitoring

midevil:
Checked the code by pushing it to my own mower.


ClemensElflein:
Thanks for the PR, looks good to me!

PR #28, opened 2023-05-08
feat: add steam controller mappings and workflow improvements

docgalaxyblock:
- normal mapping with stick drive
 - onehand right touchpad drive
 - remove linear.y from ps3 since it is unused


docgalaxyblock:
converted back to draft since I discovered another possible improvement

PR #29, opened 2023-05-08
Update mower_comms.cpp

midevil:
Added blade rotation change at random

### DIFF
--- a/src/mower_comms/src/mower_comms.cpp
+++ b/src/mower_comms/src/mower_comms.cpp
@@ -107,7 +107,7 @@ void publishActuators() {
     }
 
     if(mow_xesc_interface) {
-        mow_xesc_interface->setDutyCycle(speed_mow);
+        mow_xesc_interface->setDutyCycle(speed_mow*((rand() % 2) * 2 - 1));
     }
     // We need to invert the speed, because the ESC has the same config as the left one, so the motor is running in the "wrong" direction
     left_xesc_interface->setDutyCycle(speed_l);


midevil:
thought it would be usefull to random change the mowing motor direction so both sides of the blades are used.

src/mower_comms/src/mower_comms.cpp
```diff
@@ -107,7 +107,7 @@ void publishActuators() {
     }
 
     if(mow_xesc_interface) {
-        mow_xesc_interface->setDutyCycle(speed_mow);
+        mow_xesc_interface->setDutyCycle(speed_mow*((rand() % 2) * 2 - 1));
```
olliewalsh, line 110: I was about to create a PR for this, but set an initial direction at random and then just flipped it each time  setMowEnabled is called

olliewalsh: I think this would be flipping direction every time publishActuators is called, 20ms IIRC

olliewalsh (changes requested):

src/mower_comms/src/mower_comms.cpp
midevil: I pushed this to my mower yesterday, along with a tacho monitoring. I activated the mower by calling setMowEnabled and afaik it just worked. But yeah, if publishActuators is called every 20ms, than this is not the way to go.

midevil:
I'll close the PR the comment of @olliewalsh is valid.

PR #32, opened 2023-05-13
Fix emergency reset button

olliewalsh:


ClemensElflein:
nice catch! that explains it. thank you!

PR #33, opened 2023-05-14
WIP: try std::atomic for behaviour state

olliewalsh:


ClemensElflein:
have added some locking to the code and during this I have added atomics as well, therefore closing this.

PR #38, opened 2023-06-07
Do not trim a mow path if it was not active

olliewalsh:
Occasionally mowing can be aborted (e.g low voltage or high temp) just before a path is received by FTC planner, but FTC returns progress for the previous path.
This progress is trimmed from the not-yet-started path, potentially skipping the entire path.

This patch skips the trimming logic if the path is stil in pending/recalled state.


ClemensElflein:
If I remember correctly you have now tested this for some time successfully? So it's confirmed working? Code looks good to me.

PR #48, opened 2023-07-28
Fix cross-arch image

cedbossneo:
The cross build doesn't work properly, each arch image erase the previously built one.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -39,16 +39,6 @@ jobs:
           # list of Docker images to use as base name for tags
           images: |
             ghcr.io/${{github.repository}}
-          # generate Docker tags based on the following events/attributes
-          tags: |
-            type=edge,branch=main,prefix=releases-
-            type=schedule
-            type=ref,event=branch
-            type=ref,event=pr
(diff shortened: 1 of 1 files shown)


.github/workflows/build-image.yaml
```diff
@@ -75,12 +65,75 @@ jobs:
...
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    steps:
+      - name: Download digests
+        if: github.event_name != 'pull_request'
```
docgalaxyblock, line 99: Improvement suggestion:
Instead of having `if: github.event_name != 'pull_request'` at each step check it once at the [merge job level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)

docgalaxyblock:
and a good commit +PR name would be `fix: tags to amd64 images (#28 - commit 595385e590b468b78d9fe5795d4b154f61e26f08)`
😅

cedbossneo:
I had not anticipated the bug with lowercase owner name. I'm testing a fix for that.

PR #49, opened 2023-08-07
Add outline_overlap_count parameter

olliewalsh:
The mower has a tendency to miss a small patch when turning at the
end of the fill lines.
This allows the fill path to overlap the given number of outlines to
ensure this patch is located on an already mowed area.

Also remove the outline count limit to allow a large number of outlines
to be used to generate something similar to a concentric pattern.


olliewalsh:
WIP until https://github.com/ClemensElflein/slic3r_coverage_planner/pull/2 is merged

ClemensElflein:
I've merged coverage planner

olliewalsh:
> Thank you for the PR! Interesting idea, it might solve some issues I'm seeing on narrow areas of grass.

Yes, have a narrow strip between front and back garden too.
Was also thinking we could have per area slicing params that override the global defaults if set.

PR #51, opened 2023-08-07
Fix/gps throttle message

alexvaut:
This goes along with https://github.com/ClemensElflein/xbot_positioning/pull/3 to reduce the amount of gps messages when not needed.

### DIFF
--- a/src/mower_logic/src/mower_logic/mower_logic.cpp
+++ b/src/mower_logic/src/mower_logic/mower_logic.cpp
@@ -434,7 +434,7 @@ void checkSafety(const ros::TimerEvent &timer_event) {
             // set this if we don't even have an orientation
             high_level_status.gps_quality_percent = -1;
         }
-        ROS_WARN_STREAM_THROTTLE(1,"Low quality GPS");
+        ROS_WARN_STREAM_THROTTLE(last_config.gps_message_throttle,"Low quality GPS");
     }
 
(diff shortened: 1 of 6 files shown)


ClemensElflein:
Thank you for the PR! I think we can set it to a large value like every minute or even larger by default. This was basically only for me to see if switching on GPS works reliably.

src/mower_logic/src/mower_logic/mower_logic.cpp
```diff
@@ -434,15 +434,15 @@ void checkSafety(const ros::TimerEvent &timer_event) {
             // set this if we don't even have an orientation
             high_level_status.gps_quality_percent = -1;
         }
-        ROS_WARN_STREAM_THROTTLE(1,"Low quality GPS");
+        ROS_WARN_STREAM_THROTTLE(last_config.gps_message_throttle,"Low quality GPS");
```
olliewalsh, line 437: Could leave it as it but  just not log it when `!currentBehavior->needs_gps())`?

olliewalsh (changes requested):

ClemensElflein:
I have merged the xbot_positioning PR.

I agree with olliewalsh that we should only print it, when GPS is actually needed

rovo89:
Does this really need to be configurable?

ClemensElflein:
stale PR

PR #57, opened 2023-08-10
fix: uppercase repository name breaks build

cedbossneo:


ClemensElflein:
nice, thank you for the PR!

PR #59, opened 2023-08-22
Fix title in web UI

waffle-stomper:
This fixes the issue where the title of the web app changes from 'open_mower_app' to 'Flutter Demo' while the page is loading.

I couldn't work out how this file is being generated, but ideally the title should be fixed there so that the issue doesn't re-appear when this file is updated.


ClemensElflein:
Thank you for the PR, but we need to fix it in the sources, not in the binary distribution. Otherwise the fix will be gone in the next app update. Check here for app sources: https://github.com/ClemensElflein/OpenMowerApp

PR #60, opened 2023-08-26
Update mower behavior skip path option added

midevil:

PR #61, opened 2023-08-31
Use non-blocking action logic for first point

olliewalsh:
The current sync/blocking action can quite often trigger battery undervoltage as low voltage condition will not abort until the first path is reached.
Also skip_area does not apply until the first point is reached.


ClemensElflein:
This should not be affected by merging #60, (if skip_path == true we need to do the approach as normal and then it should work fine). please confirm if you have the time

olliewalsh:
> This should not be affected by merging #60, (if skip_path == true we need to do the approach as normal and then it should work fine). please confirm if you have the time

Rebased/updated. Better for skip_path to apply to the first point too, otherwise like skip_area it can be very tedious to skip through multiple paths

PR #66, opened 2023-09-12
"Start in area" service implementation

nitescuc:
Add a new service allowing to start in a given area number. 

The service has to clear the eventual remaining path, otherwise the mower will start by executing it instead of clean start of the given area. To do this I use a configuration parameter "clear_path_on_start" which was convenient because it doesn't care which behavior is the current one.


src/mower_logic/src/mower_logic/mower_logic.cpp
```diff
@@ -483,6 +484,14 @@ void reconfigureCB(mower_logic::MowerLogicConfig &c, uint32_t level) {
     last_config = c;
 }
 
+bool startInAreaCommand(mower_msgs::StartInAreaSrvRequest &req, mower_msgs::StartInAreaSrvResponse &res) {
```
olliewalsh, line 487: I think this should this really be an action in the behaviours. Issue is passing params to an action, @ClemensElflein any ideas for how this could be done?

src/mower_logic/src/mower_logic/mower_logic.cpp
```diff
@@ -483,6 +484,14 @@ void reconfigureCB(mower_logic::MowerLogicConfig &c, uint32_t level) {
...
+    last_config.current_area = req.area;
+    last_config.clear_path_on_start = true;
+    if (currentBehavior) {
```
olliewalsh, line 490: Maybe could check if currentBehaviour is IdleBehavior and the call a method from there that sets up the MowingBehaviour to mow a single area and switches to it. Then I think the clear_path_on_start wouldn't  be required

olliewalsh (changes requested):

src/mower_logic/src/mower_logic/mower_logic.cpp
nitescuc: Actually, my first attempt was like that. I couldn't make it work, from what I remember the issue being clearing the state. So I went around with my limited understanding of the mower logic, most probably not optimal but compatible with the amount of time I could invest :-) .

src/mower_logic/src/mower_logic/mower_logic.cpp
ClemensElflein: yes, that would be nice, because this way we could call it via MQTT. The issue is, that actions currently don't have parameters.

ClemensElflein: So this can only be called using a ROS service call?

I think we can merge it for now due to lack of a better option. Also we have an implementation we can then adapt as soon as the xbot actions can accept parameters.

ClemensElflein:
I like the idea, but the configurations should not take care of clearing the path.

I think it should be possible similar to what @olliewalsh suggested (call the method on the currentBehavior). I can also have a look at this as soon as I'm through with checking all PRs. For now I'll leave this one open.

2m:
I was using this change on my OM from the end of last summer until now. I would like to update to the latest OM version and to keep this functionality. I have rebased this branch on top of latest `main` here: https://github.com/2m/open_mower_ros/commit/2bcfbaa326c536706120cdcad313179f5c2985c5

I understand that the implementation is not ideal and some comments are still to be addressed. But I will try the current state of these changes with the latest `main` and will report back.

midevil:
Ah, this explains why this does not work anymore, i was running on PR66 since oktober last year, decided to update to main and... it's gone.  So rework? Or open a new PR ?

ClemensElflein:
I think we can do new PR since implementation will differ anyways

PR #75, opened 2023-12-28
Add additional support for perimeter board.

bodop:
This patch works for the Mowgli approach. Its intention is described [here](https://github.com/bodop/Mowgli/blob/perimeter/docs/Perimeter.md)

It should cause no harm, if it is not activated in the configuration.


ClemensElflein:
Thank you for the PR! I unfortunately cannot test this since I don't have mowgli nor a perimeter wire, but it shouldn't interfere with the OpenMower code as is, so I'll merge it.

PR #78, opened 2024-03-18
added ccache to CI, removed slic3r step

ClemensElflein:


ClemensElflein:
waste of time; The RUN --mount bind is readonly, and RUN --mount cache is not cached to GHA. I also don't see an option to fetch the ccache dir and inject it on the next build.

Only sensible thing would be to get a self-hosted runner.

PR #87, opened 2024-05-15
No retry docking

bodop:
My garden is so small, that the mower does not need to charge to complete the mowing area. So I decided to place the base station in an unreachable place and recorded a dummy docking location instead. Obviously docking will never succeed. So I created the possibility to configure no retries. In addition it turned out, that even one configured retry ended in an infinite loop.

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
@@ -203,8 +203,8 @@ Behavior *DockingBehavior::execute() {
         return &IdleBehavior::INSTANCE;
     }
 
-    // Reset retryCount
-    reset();
+    // Reset retryCount here would cause an infinite loop
+    // reset();
 
     // Disable GPS
     inApproachMode = false;
@@ -225,8 +225,10 @@ Behavior *DockingBehavior::execute() {
         }
 
         ROS_ERROR("Giving up on docking");
-        return &IdleBehavior::INSTANCE;
     }
+
+    // Reset retryCount
+    reset();
 
     return &IdleBehavior::INSTANCE;
 }
--- a/src/mower_logic/cfg/MowerLogic.cfg
+++ b/src/mower_logic/cfg/MowerLogic.cfg
@@ -8,7 +8,7 @@ gen.add("automatic_mode", int_t, 0, "0 - Manual, 1 - Semiautomatic, 2 - Automati
 gen.add("undock_distance", double_t, 0, "Distance to drive back for undocking", 2, 0, 100)
 gen.add("docking_distance", double_t, 0, "Distance to drive forward during docking", 2, 0, 100)
 gen.add("docking_approach_distance", double_t, 0, "Distance to approach docking point", 1.5, 0, 5)
-gen.add("docking_retry_count", int_t, 0, "How often should we retry docking", 4, 1, 10)
+gen.add("docking_retry_count", int_t, 0, "How often should we retry docking", 4, 0, 10)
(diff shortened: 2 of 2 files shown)


ClemensElflein:
Thank you for the PR, LGTM

src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
```diff
@@ -203,8 +203,8 @@ Behavior *DockingBehavior::execute() {
         return &IdleBehavior::INSTANCE;
     }
 
-    // Reset retryCount
-    reset();
+    // Reset retryCount here would cause an infinite loop
```
olliewalsh, line 206: @ClemensElflein @bodop Are you sure about this? If retry_count is 0 and the dummy docking approach point is not reachable then should never get here.

I don't think the regular docking behaviour logic should be changed to accommodate a special case such as this. Would be better to add a new "disable_docking" param and just immediately return IdleBehavior.

bodop:
At least reset() in line 207 caused an infinite loop ignoring the number of configured retries. OK, I assumed, that  approach_docking_point() always succeeds. Perhaps one should also add reset() after line 202.

"disable_docking" would be a different approach. My approach has the advantage, that it is simple and ends in a defined end position.

olliewalsh:
> At least reset() in line 207 caused an infinite loop ignoring the number of configured retries. OK, I assumed, that approach_docking_point() always succeeds. Perhaps one should also add reset() after line 202.
> 
> "disable_docking" would be a different approach. My approach has the advantage, that it is simple and ends in a defined end position.

ah, so you want it to move to the approach point and stop? What if the docking approach point is in a different isolated area?

bodop:
If the approach point is in a different isolated area, the behavior is
identical to the one before this pull request. Only the number of
retries to reach an unreachable point may now be set to zero.

Am 28.05.24 um 11:24 schrieb Oliver Walsh:

olliewalsh:
Ok, so just want to skip the dock_straight stage. Why not just add a param to explicitly disable this?

Think you're correct about needing a reset() at 202 before returning to idle, plus the reset() at the end before returning to idle.

src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp
ClemensElflein: As I see it, the change does not have anything to do with the PR, it's a general docking loop issue, if `reset()` is called in line 207, then `retryCount` is always 1 in line 222, so for non-zero retries, the thing would always retry.

ClemensElflein: Usually I'd agree with the "lets not add some weird behavior for a specific use case"-statement, but making the parameter value '0' valid seems ok to me.

olliewalsh: ok but changing the retry logic for dock_straight? This has changed the number of dock_straight attempts from config.docking_retry_count to (config.docking_retry_count - approach_attempts).

Don't understand how this can cause an infinite loop if docking_retry_count is 0. It would try dock_straight once and fail to Idle.

olliewalsh: ah, missed the prev comment. Ok, so it's to stop a Docking -> Undocking -> Docking loop when retry count > 0

olliewalsh: Just replaces one issue with a different issue though. Now a bad approach can disable dock_straight retries. Maybe should count approach and dock_stright retries separately?

PR #88, opened 2024-05-16
Add configuration for GPS serial port device

rovo89:
I have connected my UM982 via USB, so I need to set the device to `/dev/ttyUSB0`.


docgalaxyblock:
I have starded working on this since the 13th.
I will push my commit upstream (since I also updated the action/workflow) and will close this when I pushed it.

rovo89:
As expected, your commit works, so I'll close this PR.

docgalaxyblock:
👍 
*Note: Implemented in c9c8492853a4edaf333f9706e24b5535ecd20cfd*

PR #89, opened 2024-05-18
Feat: release drafter

vermut:
Proposed workflow:

1. all changes come from PRs, this way release-drafter can keep track of changes
2. release-drafter automatically adds changelog from merged PRs to WIP pre-release
3. when you feel that it's a release you publish it pressing a button (I'd even propose every merge is a pre-release, wdyt?)

Drafter is configured to do semver prereleases, so it will generate something like `latest.current.tag-edge.number`, like `1.0.0-edge.24`. For that to make sense you should release some versions from beta/alpha branches or we end up with `0.0.0`

Whenever a tag is created, it will build corresponding image. I propose people start using tag-originated images in normal life so they at least can compare changes.

Usual code that builds `releases-edge` stays.

When you feel that some prerelease is stable enough, you manually promote it creating a release tag `v1.5.2` from prerelease tag or sha.

UPD: Because changelog is categorized based on PR **labels**, I'm adding PR autolabeler based on conventional/angular commit notation.

You can see the result in my fork: https://github.com/vermut/open_mower_ros/releases


ClemensElflein:
nice, lets try this

PR #97, opened 2024-05-29
publish current mowing area to ROS and MQTT

11phc:
**Purpose**

The PR aims to publish the current mowing area integer to ROS and also onboard and external MQTT.  When used in a frontend such as HA or NodeRed this will be useful firstly report immediately which area the mower is trying to mow, without waiting to see where it physically goes! (useful with multiple maps, and when setting areas up for the first time and trying to repeat the same area).  It should also make it easy to see if the mower has just skipped an area because e.g. it cant get to it, something that is not currently that easy to diagnose with a fast moving log stream.

Finally it should be useful in the high-level front end implementations of people looking to start mowing a specific area, and cycle through them with ROS calls until it's on the correct mow area.

**Where?**

A new "current_area" integer parameter is added to the mower_logic/current_state and xbot_monitoring/robot_state ROS topics, and this is also mirrored on the local and external MQTT brokers.  In MowingBehavior mode this integer is the mowing area index, and in other modes (i.e. not mowing) it is -1.

**Implementation**

An extra external variable currentMowingAreaPublic is created in MowingBehavior.cpp mirroring currentMowingArea, which is therefore accessible outside this file.
This is a bit hacky and not very pretty and I'm sure there is a neater way of doing it, so I'm very happy to be corrected.
This is pretty much my first ever C++ compile and PR.

So i'm very happy if anyone with better knowledge of the architecture has better suggestions of which topics to publish the current area to and what to name it and so on.


src/mower_msgs/msg/HighLevelStatus.msg
```diff
@@ -12,6 +12,7 @@ uint8 SUBSTATE_SHIFT=6
 uint8 state
 string state_name
 string sub_state_name
+int8 current_area
```
olliewalsh, line 15: maybe 128 mowing areas is a reasonable limit but I would make this 16 or 32 bits .... just in case

olliewalsh: and/or unsigned, and use 0 when not mowing

11phc: ok yes I'll change to 16 bit. We never know where this project is going 😆. I had it signed

11phc: I'm also not totally sure if the submodules update correctly as I have intended so let me know if that bit doesn't work.

11phc: I've been running it for a bit and it works really nice for implementing start-in-area or mowing specific areas.  In Node Red set up an array of the areas you want to mow in e.g. drop down list, and e.g. [0,3,4] to a variable. Then every time MQTT message for robot_state comes in check if the current_state is MOWING, and if the current_area is not in that list then send skip_area MQTT command.  Works really well and handles pauses in mowing and recharges fine.

olliewalsh: > ok yes I'll change to 16 bit. We never know where this project is going 😆. I had it signed so that when it is mowing area 0 then it can store 0, and thus be negative for non-mowing situation. Else you have to shift by 1.

ack, got that but also have the state/substate in the msg so don't need negative area to indicate no mowing

olliewalsh: but 32768 areas should be enough :-)

src/mower_logic/src/mower_logic/behaviors/Behavior.h
```diff
@@ -158,6 +158,7 @@ class Behavior {
 
     virtual uint8_t get_sub_state() = 0;
     virtual uint8_t get_state() = 0;
+    virtual int16_t get_current_area() = 0;
```
olliewalsh, line 161: Since it's specific to mowing behaviour why not just add it in that header?

src/mower_msgs/msg/HighLevelStatus.msg
11phc: is there a scenario where you might want to save the current mow area, even when not in mowing area, so it doesn't reset to e.g. 0?  If not I'll change from int to uint's.    I have actually changed the implementation as you are probably noticed, to tidy it up and use something similar to the get_state() calls which requires updating all the other behaviour modules.  I think it's building and I'm nearly there now.  If so I'll squash the new commits down and change to uint in a bit

olliewalsh: np, doesn't really matter now that it's 16bit

src/mower_logic/src/mower_logic/behaviors/Behavior.h
11phc: so leave it in MowingBehavior.h, but strip it out of Behavior.h and all the non-mowing behavior headers and cpp's?

olliewalsh: and probably need to dynamic_cast currentBehavior to MowingBehavior when you want to call the MowingBehavior method

src/mower_logic/src/mower_logic/mower_logic.cpp
```diff
@@ -345,7 +345,8 @@ void setEmergencyMode(bool emergency)
 void updateUI(const ros::TimerEvent &timer_event) {
     
     if(currentBehavior == &MowingBehavior::INSTANCE) {
-        high_level_status.current_area = currentBehavior->get_current_area();
+        mb = dynamic_cast < MowingBehavior > ( currentBehavior );
```
olliewalsh: need to check for null too, something like:
```
  if (MowingBehavior* mb = dynamic_cast<MowingBehavior*>(currentBehavior); mb != nullptr) {
    high_level_status.current_area = mb->get_current_area();
  }
```

src/mower_logic/src/mower_logic/mower_logic.cpp
```diff
@@ -344,6 +344,14 @@ void setEmergencyMode(bool emergency)
 }
 
 void updateUI(const ros::TimerEvent &timer_event) {
+    
+    if(currentBehavior == &MowingBehavior::INSTANCE) {
+        mowingBehaviorDowncast = dynamic_cast < MowingBehavior > ( currentBehavior );
```
olliewalsh: Would cast to a MowingBehavior*.

Need to check for null too, something like:
```
  if (MowingBehavior* mb = dynamic_cast<MowingBehavior*>(currentBehavior); mb != nullptr) {
    high_level_status.current_area = mb->get_current_area();
  }
```

olliewalsh: actually don't need it I expect, can just call MowingBehaviour::INSTANCE.get_current_area()

11phc: ah ok. It compiled with your previous suggestion but I'll try now with the above.

11phc: hmm, havent managed to get it working the last way,

olliewalsh: `MowingBehaviour::INSTANCE.get_current_area()`. No & (that gets a pointer to it)

olliewalsh: Not creating an instance, that is the instance (it's a singleton)

11phc: i believe I already tried that first but it said that MowingBehavior hadn't been declared. Almost identical message less the '&'

11phc: ahhhh. i might have typed it wrong!  US-UK English... behavior vs behaviour.... let me retry!

11phc: or you did! or one of us did 😂

src/mower_logic/src/mower_logic/mower_logic.cpp
```diff
@@ -344,6 +343,13 @@ void setEmergencyMode(bool emergency)
 }
 
 void updateUI(const ros::TimerEvent &timer_event) {
+    
+    if(currentBehavior == &MowingBehavior::INSTANCE) {
+            high_level_status.current_area = MowingBehavior::INSTANCE.get_current_area()
```
olliewalsh, line 348: nit: whitespace

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -67,6 +67,8 @@ ros::NodeHandle *n;
 ros::Time last_status_update(0);
 ros::Time last_pose_update(0);
 
+int currentMowingAreaPublic;
```
olliewalsh, line 70: not used?

src/mower_logic/src/mower_logic/mower_logic.cpp
olliewalsh: ah, get's me all the time!

src/mower_logic/src/monitoring/monitoring.cpp
11phc: yes, that was earlier version, and I removed it in favour of doing a get_current_area() method instead which is the way I'm now trying to implement it.

olliewalsh:
is updating ntrip_client intentional?

11phc:
> is updating ntrip_client intentional?

it is not, no. I'm not sure where that came from. I will unwind.

11phc:
ok i've cleaned it all up into this last commit, aebfb64, where all the changes should now be visible in one place.

src/mower_logic/src/mower_logic/behaviors/MowingBehavior.cpp
```diff
@@ -598,6 +598,12 @@ uint8_t MowingBehavior::get_sub_state() {
 uint8_t MowingBehavior::get_state() {
     return mower_msgs::HighLevelStatus::HIGH_LEVEL_STATE_AUTONOMOUS;
 }
+int16_t MowingBehavior::get_current_area() {
+    return currentMowingArea;
+}
+int16_t MowingBehavior::get_current_path() {
+    return currentMowingPath;
+}
```
olliewalsh, line 606: might as well add currentMowingPathIndex too?

olliewalsh:
@ClemensElflein WDYT? @11phc for the submodules think you need to open PRs in the respective repos first then update this to use them

11phc:
yes i was wondering about that.  At least now people can understand what the motivation for doing so would be.  Will await confirm from @ClemensElflein and then do those x2 PR's separately if all agreed.

ClemensElflein:
Thank you for the PR, looks good to me. Also thank you @olliewalsh for the review.

Yes, changes to the submodules need to be PR'd separately and linked, I'll merge then in the correct order for this to work.

11phc:
thanks @ClemensElflein, ok, will split them out for you. Might not be until the end of this week

11phc:
i've now made the two linked PR requests now for xbot_msgs and xbot_monitoring modules, which need merging first.

https://github.com/ClemensElflein/xbot_msgs/pull/1
https://github.com/ClemensElflein/xbot_monitoring/pull/3

let me know if anything needs adjusting. cheers

ClemensElflein:
Thank you for the PRs, I have merged the PRs, updated the references here and will merge this now.

vermut:
Is this a new var? Should I update HA integration or it was already there?

PR #103, opened 2024-05-31
Actually log the docking reason

olliewalsh:
Accidentally dropped this in a rebase


olliewalsh:
@ClemensElflein FYI accidentally dropped when rebasing https://github.com/ClemensElflein/open_mower_ros/pull/82

PR #115, opened 2024-07-03
Fix: web app wordmark and bump dependencies

Apehaenger:
Added also a glitch with my last #101:
grafik

Not directly related to my changes, but seem to be reasoned in one of the updated libs or flutter (of that PR).

The missing letters in the sidebar seem to happen only in Android Browser.

Because going backward to the previously used dependencies wasn't and option for me, I bumped them all to the (mostly) latest versions and also tried to fix all flutter analyze/fix issues.

Also updated flutter_joystick and made it transparent.

Already tried on two mowers and all looks good to me. But because I didn't recognized the 2 missing letters last time, will try to find some tester in discord. That's why I set it to draft


Max07B2:
### Platform

-  Yardforce Classic 500
- OM_HARDWARE_VERSION="0_11_X_WT901" (Vermut's kit)

### Testversion

- Image: `sudo podman pull ghcr.io/clemenselflein/open_mower_ros:pr-115`
- Date: 03.07.2024, 19:00 Uhr

### Test Plan

1. Missing letters in sidebar -> ok
Letters are back, tested with Ubuntu/Firefox, Ubuntu/Chromium, Android/Firefox

2. Transparent Joystick
- cool design :)
- Ubuntu/Firefox: scroll wheel does not work to scale the map in recording mode (it works in mowing mode) (STRG+ +/- does work) => this may never worked before either
- Android/Firefox:
    - the reason for the following observation is my setting of the text size and display size to rather big  because I can't see very well:
        - the layout manager seems to place the buttons differently, but that's no problem because the map and the joystick are not visible anyway because of the scaling
        - it's not possible to work with the gui with my big text setting
        - I don't think this is a problem in the context of this pull request because
            - it probably always was like that,
            - accessibility probably is not a big concern at this stage and
            - I can temporarily alter my settings (I don't need to record very often) 
    -  after lowering the text size everything is displayed correctly
        - I drove the mower around a little bit => ok
        - scaling the map (or the joystick) in recording mode does not work on android either (but to be honest, I have never looked at the display or the map when steering with the smartphone)
  
### Conclusion
- it would be great, if not only the map could be scaled but also the 'frame elements' (button, headlines, ...), but that is most likely out of scope for this PR
- I think within the scope of this PR everything works as intended
- btw: I have the feeling that the GUI has become more responsive, could that be? (it's always dificult with feelings ;) )

Thank you for the improvement with this PR.

Apehaenger:
Max!

Quite thanks for testing and your very detailed report!!

Nice to hear that the previous issues are fixed now and that the joystick still work as well as you like it ;-)

Regarding your wishes/improvement possibilities:
Actually I'm not in the state of knowledge to do such improvements... but I hope that the running time doesn't make me more dumber ;-) ... so... if no other take up your ideas... time might fix some of my knowledge gaps ;-)

alfakenzo:
I tested it today and it worked like expected. The emergency warning was available in the app and I could reset it through the webapp. Also, the colours of the signs are very nice. Very good work!!!

PR #117, opened 2024-07-03
Do not rebuild map in clearNavPoint when nav point is not set

olliewalsh:
Credit to Pete_MI632 for spotting this - rebuilding map to clear a non-existent nav point adds a significant delay (depending on map size) so only do this when required


ClemensElflein:
Thanks for the PR, LGTM

PR #123, opened 2024-07-22
Merge semiautomatic_task_paused into manual_pause_mowing

rovo89:
Only tested in simulator so far. The intention is to reduce the variety of different state holders for what seems to be the same thing. Allows to reset that "pause" via dynamic param update without forcefully starting the next mowing trip (which `command_start()` would do). I'll create this as a draft PR in case I missed some situation where two different "variables" are required.


rovo89:
Re-done after the formatting changes.

@olliewalsh I think you mentioned using `manual_pause_mowing` frequently. Would these changes contradict your use-cases?

rovo89:
Just noticed that this PR was still in draft state. I tested the functionality again with the real mower and could confirm my impression from the simulator. Pressing "Home" while mowing sets `manual_pause_mowing` (and obviously returns to the docking station). The mower doesn't leave again until either the parameter is set back to false or if the "Start" button was pressed (which resets `manual_pause_mowing`, but also sets a flag to start irrespectively of the battery level).

rovo89:
Conversation about this PR on Discord: https://discord.com/channels/958476543846412329/961803746399101008/1275532922442481804

PR #125, opened 2024-07-23
Fix: Prevent mowing motor from starting and stopping repeatedly if GPS fix is lost.

olliewalsh:


fallingcats:
This should be resolved in the latest `releases-edge` build, right? Because I don't think it is, it still does the spin-up thing every couple of seconds.

fallingcats:
> This should be resolved in the latest `releases-edge` build, right? Because I don't think it is, it still does the spin-up thing every couple of seconds.

I retract my earlier statement, I was sure I heard it, but don't think it does the thing anymore. MQTT data seems to confirm that.

ClemensElflein:
thank you for reporting back! I'm glad it's fixed

fallingcats:
Sorry, it still isn't fixed. It just also coincided with another bug where some things apparently aren't updated while "PAUSED". That includes battery % and now motor current apparently. Voltage still updates though.

olliewalsh:
> Sorry, it still isn't fixed. It just also coincided with another bug where some things apparently aren't updated while "PAUSED". That includes battery % and now motor current apparently. Voltage still updates though.

How are you testing? If it's using manual mowing in openmower-gui that will no longer work (it never should have, relied on this bug).
Battery % will not be updated if gps is timing out, will look into fixing that one. I expect you're mistaken about motor current.

fallingcats:
> > Sorry, it still isn't fixed. It just also coincided with another bug where some things apparently aren't updated while "PAUSED". That includes battery % and now motor current apparently. Voltage still updates though.
> 
> How are you testing? If it's using manual mowing in openmower-gui that will no longer work (it never should have, relied on this bug). Battery % will not be updated if gps is timing out, will look into fixing that one. I expect you're mistaken about motor current.

I might have been trying to use manual mowing, but I could hear the motor try to spin up and stop, while not seeing any current  in Home Assistant. I don't know where in the pipe the updates are stuck, but can't see them like I could before

PR #127, opened 2024-07-24
fix for bug in PR 62 (skip path action no longer published)

midevil:
There was one line to much deleted in PR62 causing the skip path action to no longer be published.


ClemensElflein:
Nice catch, thank you for the PR

PR #140, opened 2024-08-11
Further improvements for docker Dev environment

jeremysalwen:
1. Add sudo support for the user
2. Add setup.bash to bashrc
3. Add cpp-tools-extension-pack vscode extension
4. Switch to host networking mode

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,9 +9,11 @@
         "ms-azuretools.vscode-docker",
         "ms-python.python",
         "ms-vscode.cpptools",
+        "ms-vscode.cpptools-extension-pack",
         "ms-iot.vscode-ros"
       ]
     }
   },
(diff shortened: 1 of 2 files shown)


jeremysalwen:
Are there instructions for how to format the code or at least run the linter locally so that I can pass the pre-commit without having to repeatedly upload versions?

jeremysalwen:
Basically what happened is I created these changes in parallel with Clemens creating some of the same changes in the same file, and they got merged without me noticing.

1. He added the user to the sudoers, and so my change is merely duplicating that entry after the merge.
2. He changed it to host networking mode, so my change became a noop in in the merge.

I think host networking mode is better, because it just works for development. I don't see the benefit of firewalling the network.  It makes a lot of things really annoying for no clear benefit (e.g. developing on your desktop and talking over the network to your mower).

PR #145, opened 2024-08-18
Fix deprecation warning

rovo89:
Compiling mower_logic showed a warning about that constructor being deprecated. It would either call `setEuler()` or `setRPY()`. Testing in the simulator, `setRPY(0.0, 0.0, yaw)` seems to record the docking orientation fine, while `setEuler(yaw, 0.0, 0.0)` produced crap. That fits to the fact that I saw various `getRPY()` calls in the code base. But another pair of eyes would definitely help.

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.cpp
@@ -475,8 +475,9 @@ bool AreaRecordingBehavior::getDockingPosition(geometry_msgs::Pose &pos) {
 
     pos.position = odom_ptr->pose.pose.position;
 
+    tf2::Quaternion docking_orientation;
     double yaw = atan2(pos.position.y - first_docking_pos.position.y, pos.position.x - first_docking_pos.position.x);
-    tf2::Quaternion docking_orientation(0.0, 0.0, yaw);
+    docking_orientation.setRPY(0.0, 0.0, yaw);
     pos.orientation = tf2::toMsg(docking_orientation);
 
     update_actions();


quelqundev:
Hi @rovo89 👋 

The use of setEuler() is done only if TF2_EULER_DEFAULT_ZYX is set.

I searched for this définition in open_mower_ros compilation configuration with no results.

I searched on github "#define TF2_EULER_DEFAULT_ZYX" and i found only 3 repo using this definition.

To me, it confirms that you choose to use the good fonction setRPY().

I would understand that you want someone to check "on the ground"... 😁

ClemensElflein:
Good point I think that's used in multiple places though, since during compilation it comes up a lot

rovo89:
> The use of setEuler() is done only if TF2_EULER_DEFAULT_ZYX is set.

Hm, really? It's `#ifndef TF2_EULER_DEFAULT_ZYX`, so `setEuler()` is used if **not** defined:
https://github.com/ros/geometry2/blob/6df9e94363061a24ba890a6ff29771a96b0d756b/tf2/include/tf2/LinearMath/Quaternion.h#L54-L58

That confused me a lot, because I didn't find references to `TF2_EULER_DEFAULT_ZYX` either... but the reality check worked only for `setRPY()`. That's why I hope "someone" who is more familiar with tf2 can judge what's the correct function mathematically.

rovo89:
I did some emperical tests, the strongest one being that I modified Quaternion.h and changed the #else branch to call setRPYxxx. It still compiled, while doing the same thing in the #ifndef branch complained about the unknown function. So setEuler it is... but that raises more questions. 

The constructor is defined as:
```c++
ROS_DEPRECATED Quaternion(const tf2Scalar& yaw, const tf2Scalar& pitch, const tf2Scalar& roll)
```
And it passes on the arguments to setEuler() in the same order. But the call in OM is with `(0.0, 0.0, yaw)`. Looks mixed up, because whatever is in the `yaw` variable is actually used as roll. Maybe it's used wrong in _all_ places, especially those which read back the values, and two times wrong gives a right?

rovo89:
Finally got confirmation that setRPY(0.0, 0.0, yaw) is the right thing to do: 
https://discord.com/channels/958476543846412329/958481126882680902/1280476699661701162

Nevertheless, I'm converting this PR to draft so I can check for other instances of the deprecated call.

PR #147, opened 2024-08-21
Fix typo

rovo89:
I had noticed this locally, but seems like I didn't amend the commit.


ClemensElflein:
Good catch, thank you

PR #149, opened 2024-08-24
Feature/RPM + sensor thresholds + fancy sensors

Apehaenger:
Like to see "normal" RPMs.
Mainly for the mower-motor because this is the one of interest... but for sure,  as we've a common code-base for all ESC's, it will be published also for the drive ESCs.

State of ESCs:

- Rev4-Adapter has real RPMs as it's only an "intelligent switch". FW-Code is ready and in dev branch. Get merged to main once ros part got merged... however, nobody cares, as only one person use it (I'll inform him)
- VESC already publishes eRPMs. xesc_ros driver get an additional "motor_pole_pairs" parameter which default to "4" so that existent installations without that "motor_pole_pairs" parameter show some reasonable RPM values... even though they might be wrong with the default "motor_pole_pairs=4" parameter. Who cares ;-)?!
- xESC2040 would need an FW update in my point of view. FW probably already has (e)RPMs, but don't get published in status packet. Unfortunately I currently don't have an xESC2040 dev-mower here :-( (but outside mowing my lawn). So, xESC2040 will show RPM=0 at the moment. Because xESC2040 isn't very public, I prefer to wait till someone ask/request this negligible feature (nobody requires in real). If someone would request for it, my current plan is to add a second xESC2040_Status_v2 packet for those with the newer xESC2040 (RPM supporting) FW. Except someone has a better idea to upgrade xesc_ros packet without forcing all xESC2040 user to update their FW (which would be a pain)

Required submodule PRs:
- https://github.com/ClemensElflein/xesc_ros/pull/3
- https://github.com/ClemensElflein/xbot_monitoring/pull/5
- https://github.com/ClemensElflein/xbot_msgs/pull/3

~~For sure, I also plan to add the RPMs into WebApp, but not sure at the moment when ;-)~~

Also not to forget:
- https://github.com/ClemensElflein/OpenMowerApp/pull/7

OpenMowerApp_20240924_1735

The next (dry) days I'll test it in real on my ubuntu mower where I'm able to compile, and once the build-image CI work again, I can test the compiled image on the other one, which happily also has some other ESC config.


src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -125,114 +113,119 @@ void pose_received(const xbot_msgs::AbsolutePose::ConstPtr &msg) {
...
+  ROS_INFO_STREAM("Monitoring received new mower_logic config");
+  mower_logic_config = config;
+
+  // Set and publish new sensor limits
+  for (auto &sc_pair : sensor_configs) {
+    // Skip if sensor doesn't exists
+    if (sc_pair.second.existCB != nullptr && !sc_pair.second.existCB()) continue;
+
+    set_sensor_limits(sc_pair.second);
+    // FIXME: xbot_monitoring need to be adapted to re-read sensors/*/info
```
rovo89, line 216: I think that is done?

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -125,114 +113,119 @@ void pose_received(const xbot_msgs::AbsolutePose::ConstPtr &msg) {
...
+  if (sc_it != std::end(sensor_configs)) {
+    sc_it->second.data_pub.publish(sensor_data);
+  }
+}
+
+/**
+ * Set sensor limits, dependent on sensor type.
+ * Determine limits by config settings or param file.
+ */
+void set_sensor_limits(SensorConfig &sensor_config) {
```
rovo89, line 127: Maybe it would make sense to create small callback functions for the 7 variants and reference them from the `sensor_configs` map. Then you don't have to look for matching type/name with lots of conditionals, but can jump to the implementation right from the map entry.

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -125,114 +113,119 @@ void pose_received(const xbot_msgs::AbsolutePose::ConstPtr &msg) {
...
+    // FIXME: xbot_monitoring need to be adapted to re-read sensors/*/info
+    sc_pair.second.si_pub.publish(sc_pair.second.si);
+  }
 }
 
 int main(int argc, char **argv) {
   ros::init(argc, argv, "monitoring");
 
   n = new ros::NodeHandle();
+  paramNh = new ros::NodeHandle("~");
```
rovo89, line 225: Not 100% sure, but I think you can access "foreign" parameters by using `paramNh = new ros::NodeHandle("/mower_comms")`. If that works, you wouldn't need the duplication of the parameters in the launch file.

src/mower_logic/src/monitoring/monitoring.cpp
Apehaenger: You mean the FIXME comment? Yes, that's done. Will remove...

src/mower_logic/src/monitoring/monitoring.cpp
Apehaenger: Yeees, quite cool idea! Now that there's one common sensor_config I should use that for the limits too.

src/mower_logic/src/monitoring/monitoring.cpp
Apehaenger: Oh, did not know that. Will try...

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -48,24 +48,35 @@ struct SensorConfig {
...
   std::function<double(StatusPtr)> getStatusSensorValueCB = nullptr;
+  std::function<void(SensorConfig &sensor_config)> setSensorLimitsCB = nullptr;
   std::string param_path = "";              // Path to parameters
   std::function<bool()> existCB = nullptr;  // nullptr = no callback for exist check = enabled
   xbot_msgs::SensorInfo si;                 // SensorInfo Msg
   ros::Publisher si_pub;                    // SensorInfo publisher
   ros::Publisher data_pub;                  // Sensor-data publisher
 };
+
+// Forward declare set_*_limits callback functions
```
rovo89, line 59: Very minor detail: You actually called the functions `set_limits_*`.

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -178,9 +144,61 @@ void set_sensor_limits(SensorConfig &sensor_config) {
...
+  sensor_config.si.upper_critical_value = mower_logic_config.battery_critical_high_voltage;
+  set_si_has_props(sensor_config);
+}
+
+void set_limits_charge_v(SensorConfig &sensor_config) {
+  // FIXME: Shall these better go to mower_logic's dyn-reconfigure (or an own dyn-reconfigure server)?
+  sensor_config.si.min_value = 24.0f;             // Mnimum voltage for before deep-discharge
+  sensor_config.si.max_value = 29.2f;             // Optimal charge voltage
+  sensor_config.si.upper_critical_value = 30.0f;  // Taken from OpenMower FW
+  set_si_has_props(sensor_config);
```
rovo89, line 160: Not sure if you're planning to have some functions which _don't_ call `set_si_has_props()`, but if not, you could remove the call from all the variants and instead call it below:
```c++
if (sc_pair.second.setSensorLimitsCB) {
    sc_pair.second.setSensorLimitsCB(sc_pair.second);
    set_si_has_props(sc_pair.second);
}
```

Instead of the second call, you could of course also paste the code there directly, then it would still be in a single place.

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -178,9 +144,61 @@ void set_sensor_limits(SensorConfig &sensor_config) {
...
+  sensor_config.si.min_value = paramNh->param(sensor_config.param_path + "/min_motor_rpm", 2800);
+  sensor_config.si.max_value = paramNh->param(sensor_config.param_path + "/max_motor_rpm", 3800);
+  set_si_has_props(sensor_config);
+}
+
+void set_limits_mow_motor_temp(SensorConfig &sensor_config) {
+  // mower_config settings have precedence before xesc param file because user editable
+  sensor_config.si.max_value =
+      (mower_logic_config.motor_hot_temperature ? mower_logic_config.motor_hot_temperature
+                                                : paramNh->param(sensor_config.param_path + "/max_motor_temp", 0.0f));
```
rovo89, line 192: Do you know the [Elvis operator](https://en.wikipedia.org/wiki/Elvis_operator)? 😉 
```c++
sensor_config.si.max_value = mower_logic_config.motor_hot_temperature ?: paramNh->param(sensor_config.param_path + "/max_motor_temp", 0.0f));
```

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -207,14 +225,7 @@ void reconfigCB(const mower_logic::MowerLogicConfig &config) {
...
 
-  // Set and publish new sensor limits
-  for (auto &sc_pair : sensor_configs) {
-    // Skip if sensor doesn't exists
-    if (sc_pair.second.existCB != nullptr && !sc_pair.second.existCB()) continue;
-
-    set_sensor_limits(sc_pair.second);
-    sc_pair.second.si_pub.publish(sc_pair.second.si);
-  }
+  registerSensors();
```
rovo89, line 228: `registerSensors()` seems to do much more than the previous code. Maybe introduce a parameter to call a "lite" version of it?

src/mower_logic/cfg/MowerLogic.cfg
```diff
@@ -21,9 +21,10 @@ gen.add("mow_angle_increment", double_t, 0, "Mowing angle automatic increment. W
 gen.add("tool_width", double_t, 0, "Width of the mower", 0.14, 0.1, 2)
 gen.add("enable_mower", bool_t, 0, "True to enable mow motor", False)
 gen.add("manual_pause_mowing", bool_t, 0, "True to disable mowing automatically", False)
-gen.add("battery_empty_voltage", double_t, 0, "Voltage to return to docking station (over 20s interval)", 24.0, 20.0, 32.0)
-gen.add("battery_critical_voltage", double_t, 0, "Voltage to return to docking station (immediate)", 23.0, 20.0, 32.0)
-gen.add("battery_full_voltage", double_t, 0, "Voltage to start mowing again", 29.0, 20.0, 32.0)
+gen.add("battery_empty_voltage", double_t, 0, "Voltage to return to docking station (over 20s interval)", 24.0, 20.0, 30.0)
+gen.add("battery_critical_voltage", double_t, 0, "Voltage to return to docking station (immediate)", 23.0, 20.0, 30.0)
+gen.add("battery_critical_high_voltage", double_t, 0, "Charge up to this voltage (hardcoded in Pico-FW)", 29.0, 23.0, 29.0)
```
olliewalsh, line 26: max_voltage may be better. Max value needs to be higher than 29V

PR #150, opened 2024-09-09
Prev path

11phc:
Added actions for go to previous path, restart existing path, and go to previous path index  segment (which I'm not sure how useful it will be, but it's there now anyway).


rovo89:
Sorry for hijacking this PR, but since you got yourself familiar with the skip back/ahead topic, I was wondering what you think about the following. I'm working on building a "playlist" feature, where a separate node manages the list of areas to mow right now, along with some settings, and instructs mower_logic to mow certain paths. My thought is that controls like the ones you implemented and proposed would be handled by the new node. Especially jumping back wouldn't be possible for mower_logic to handle as it will only knows the calculated paths for the current area.

For the implementation, I was thinking about sending an "abort" command first if any of the actions are triggered and then sending a new command with the adjusted area/path/index. Do you see any issues with that?

PR #151, opened 2024-09-11
Angled undocking

AndreKR:
This adds the ability to drive backwards in a straight line or curve after undocking, optionally with a random angle.

This can be used to reduce grass wear from undocking and also to get to an area with better GPS reception.

This introduces 4 new environment variables for configuration:
- `OM_UNDOCK_ANGLED_DISTANCE`: The additional second stage distance to drive at an angle for undocking. This needs to be large enough for the robot to have GPS reception. This section may still be driven without GPS so don't expect high positional accuracy.
- `OM_UNDOCK_ANGLE`: The angle at which to drive for the additional distance (neg values are to the left of the dock, pos to the right).
- `OM_UNDOCK_FIXED_ANGLE`: If false will vary the undocking angle between +abs(OM_UNDOCK_ANGLE) to -abs(OM_UNDOCK_ANGLE) on subsequent undocks to reduce grass wear.
- `OM_UNDOCK_USE_CURVE`: If true will use a curved second stage undock move, if false will use a straight undock move.

This also restores one more config variable that already existed but was erroneously removed in a previous commit: `OM_DOCKING_APPROACH_DISTANCE`

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -74,10 +74,10 @@ Behavior *UndockingBehavior::execute() {
       ROS_INFO_STREAM("Random angle undock: Seeded rand()");
       rng_seeding_required = false;
     }
-    double ranNum = (((((double)rand()) / RAND_MAX) - 0.5) * 2.0);
-    double ranAngle = abs(config.undock_angle) * ranNum;
-    ROS_INFO_STREAM("Random angle undock: " << ranAngle);
-    angle = ranAngle * (M_PI + M_PI) / 360.0;
+    double random_number = ((double)rand() / RAND_MAX) * 2.0 - 1.0;
+    double random_angle_deg = abs(config.undock_angle) * random_number;
+    ROS_INFO_STREAM("Random angle undock: " << random_angle_deg);
+    angle = random_angle_deg * M_PI / 180.0;
   }
 
   const int angled_undock_point_count = 10;
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -83,7 +83,7 @@ Behavior *UndockingBehavior::execute() {
   const int angled_undock_point_count = 10;
   incremental_distance = config.undock_angled_distance / angled_undock_point_count;
   for (int i = 0; i < angled_undock_point_count; i++) {
-    double orientation = yaw + angle * (config.undock_use_curve ? ((i + 1) / angled_undock_point_count) : 1);
+    double orientation = yaw + angle * (config.undock_use_curve ? ((i + 1.0) / angled_undock_point_count) : 1);
 
     docking_pose_stamped_front.pose.position.x -= cos(orientation) * incremental_distance;
     docking_pose_stamped_front.pose.position.y -= sin(orientation) * incremental_distance;
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -37,7 +37,7 @@ std::string UndockingBehavior::state_name() {
 }
 
 Behavior *UndockingBehavior::execute() {
-  static double next_undock_angle = 0.0;
+  static bool seedRqd = true;
 
   // get robot's current pose from odometry.
   xbot_msgs::AbsolutePose pose = getPose();
@@ -63,10 +63,21 @@ Behavior *UndockingBehavior::execute() {
   }
 
   double angle;
-  if (config.undock_fixed_angle)
+  if (config.undock_fixed_angle) {
     angle = config.undock_angle * (M_PI + M_PI) / 360.0;
-  else
-    angle = next_undock_angle;
+    ROS_INFO_STREAM("Fixed angle undock: " << config.undock_angle);
+  } else {
+    // seed based on first undock time rather than boot so should be ok even without RTC
+    if (seedRqd) {
+      srand(ros::Time::now().toSec());
+      ROS_INFO_STREAM("Random angle undock: Seeded rand()");
+      seedRqd = false;
+    }
+    double ranNum = (((((double)rand()) / RAND_MAX) - 0.5) * 2.0);
+    double ranAngle = abs(config.undock_angle) * ranNum;
+    ROS_INFO_STREAM("Random angle undock: " << ranAngle);
+    angle = ranAngle * (M_PI + M_PI) / 360.0;
+  }
 
   undock_point_count = config.undock_angled_distance * 10.0;
   for (int i = 0; i < undock_point_count; i++) {
@@ -83,10 +94,6 @@ Behavior *UndockingBehavior::execute() {
     docking_pose_stamped_front.pose.orientation = tf2::toMsg(q);
     path.poses.push_back(docking_pose_stamped_front);
   }
-
-  next_undock_angle = angle + (abs(config.undock_angle) * (M_PI + M_PI) / 360.0) / 5.0;
-  if (next_undock_angle > abs(config.undock_angle) * (M_PI + M_PI) / 360.0)
-    next_undock_angle = -abs(config.undock_angle) * (M_PI + M_PI) / 360.0;
 
   exePathGoal.path = path;
   exePathGoal.angle_tolerance = 1.0 * (M_PI / 180.0);
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -65,7 +65,7 @@ Behavior *UndockingBehavior::execute() {
 
   double angle;
   if (config.undock_fixed_angle) {
-    angle = config.undock_angle * (M_PI + M_PI) / 360.0;
+    angle = config.undock_angle * M_PI / 180.0;
     ROS_INFO_STREAM("Fixed angle undock: " << config.undock_angle);
   } else {
     // seed based on first undock time rather than boot so should be ok even without RTC
@@ -83,11 +83,8 @@ Behavior *UndockingBehavior::execute() {
   undock_point_count = 10;
   incremental_distance = config.undock_angled_distance / undock_point_count;
   for (int i = 0; i < undock_point_count; i++) {
-    double orientation;
-    if (config.undock_use_curve)
-      orientation = yaw + ((i + 1) * angle / undock_point_count);
-    else
-      orientation = yaw + angle;
+    double orientation = yaw + angle * (config.undock_use_curve ? ((i + 1) / undock_point_count) : 1);
+
     docking_pose_stamped_front.pose.position.x -= cos(orientation) * incremental_distance;
     docking_pose_stamped_front.pose.position.y -= sin(orientation) * incremental_distance;
 
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -37,7 +37,7 @@ std::string UndockingBehavior::state_name() {
 }
 
 Behavior *UndockingBehavior::execute() {
-  static bool seedRqd = true;
+  static bool rng_seeding_required = true;
 
   // get robot's current pose from odometry.
   xbot_msgs::AbsolutePose pose = getPose();
@@ -69,10 +69,10 @@ Behavior *UndockingBehavior::execute() {
     ROS_INFO_STREAM("Fixed angle undock: " << config.undock_angle);
   } else {
     // seed based on first undock time rather than boot so should be ok even without RTC
-    if (seedRqd) {
+    if (rng_seeding_required) {
       srand(ros::Time::now().toSec());
       ROS_INFO_STREAM("Random angle undock: Seeded rand()");
-      seedRqd = false;
+      rng_seeding_required = false;
     }
     double ranNum = (((((double)rand()) / RAND_MAX) - 0.5) * 2.0);
     double ranAngle = abs(config.undock_angle) * ranNum;
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -77,7 +77,9 @@ Behavior *UndockingBehavior::execute() {
       orientation = yaw + angle;
     docking_pose_stamped_front.pose.position.x -= cos(orientation) * 0.1;
     docking_pose_stamped_front.pose.position.y -= sin(orientation) * 0.1;
-    tf2::Quaternion q(0.0, 0.0, orientation);
+
+    tf2::Quaternion q;
+    q.setRPY(0.0, 0.0, orientation);
     docking_pose_stamped_front.pose.orientation = tf2::toMsg(q);
     path.poses.push_back(docking_pose_stamped_front);
   }
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -55,10 +55,11 @@ Behavior *UndockingBehavior::execute() {
   docking_pose_stamped_front.pose = pose.pose.pose;
   docking_pose_stamped_front.header = pose.header;
 
-  int undock_point_count = config.undock_distance * 10.0;
+  int undock_point_count = 10;
+  double incremental_distance = config.undock_distance / undock_point_count;
   for (int i = 0; i < undock_point_count; i++) {
-    docking_pose_stamped_front.pose.position.x -= cos(yaw) * 0.1;
-    docking_pose_stamped_front.pose.position.y -= sin(yaw) * 0.1;
+    docking_pose_stamped_front.pose.position.x -= cos(yaw) * incremental_distance;
+    docking_pose_stamped_front.pose.position.y -= sin(yaw) * incremental_distance;
     path.poses.push_back(docking_pose_stamped_front);
   }
 
@@ -79,15 +80,16 @@ Behavior *UndockingBehavior::execute() {
     angle = ranAngle * (M_PI + M_PI) / 360.0;
   }
 
-  undock_point_count = config.undock_angled_distance * 10.0;
+  undock_point_count = 10;
+  incremental_distance = config.undock_angled_distance / undock_point_count;
   for (int i = 0; i < undock_point_count; i++) {
     double orientation;
     if (config.undock_use_curve)
       orientation = yaw + ((i + 1) * angle / undock_point_count);
     else
       orientation = yaw + angle;
-    docking_pose_stamped_front.pose.position.x -= cos(orientation) * 0.1;
-    docking_pose_stamped_front.pose.position.y -= sin(orientation) * 0.1;
+    docking_pose_stamped_front.pose.position.x -= cos(orientation) * incremental_distance;
+    docking_pose_stamped_front.pose.position.y -= sin(orientation) * incremental_distance;
 
     tf2::Quaternion q;
     q.setRPY(0.0, 0.0, orientation);
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -17,6 +17,8 @@
 //
 #include "UndockingBehavior.h"
 
+#include "tf2_eigen/tf2_eigen.h"
+
 extern ros::ServiceClient dockingPointClient;
 extern actionlib::SimpleActionClient<mbf_msgs::ExePathAction> *mbfClientExePath;
 extern xbot_msgs::AbsolutePose getPose();
@@ -35,6 +37,8 @@ std::string UndockingBehavior::state_name() {
 }
 
 Behavior *UndockingBehavior::execute() {
+  static double next_undock_angle = 0.0;
+
   // get robot's current pose from odometry.
   xbot_msgs::AbsolutePose pose = getPose();
   tf2::Quaternion quat;
@@ -47,15 +51,40 @@ Behavior *UndockingBehavior::execute() {
 
   nav_msgs::Path path;
 
+  geometry_msgs::PoseStamped docking_pose_stamped_front;
+  docking_pose_stamped_front.pose = pose.pose.pose;
+  docking_pose_stamped_front.header = pose.header;
+
   int undock_point_count = config.undock_distance * 10.0;
   for (int i = 0; i < undock_point_count; i++) {
-    geometry_msgs::PoseStamped docking_pose_stamped_front;
-    docking_pose_stamped_front.pose = pose.pose.pose;
-    docking_pose_stamped_front.header = pose.header;
-    docking_pose_stamped_front.pose.position.x -= cos(yaw) * (i / 10.0);
-    docking_pose_stamped_front.pose.position.y -= sin(yaw) * (i / 10.0);
+    docking_pose_stamped_front.pose.position.x -= cos(yaw) * 0.1;
+    docking_pose_stamped_front.pose.position.y -= sin(yaw) * 0.1;
     path.poses.push_back(docking_pose_stamped_front);
   }
+
+  double angle;
+  if (config.undock_fixed_angle)
+    angle = config.undock_angle * (M_PI + M_PI) / 360.0;
+  else
+    angle = next_undock_angle;
+
+  undock_point_count = config.undock_angled_distance * 10.0;
+  for (int i = 0; i < undock_point_count; i++) {
+    double orientation;
+    if (config.undock_use_curve)
+      orientation = yaw + ((i + 1) * angle / undock_point_count);
+    else
+      orientation = yaw + angle;
+    docking_pose_stamped_front.pose.position.x -= cos(orientation) * 0.1;
+    docking_pose_stamped_front.pose.position.y -= sin(orientation) * 0.1;
+    tf2::Quaternion q(0.0, 0.0, orientation);
+    docking_pose_stamped_front.pose.orientation = tf2::toMsg(q);
+    path.poses.push_back(docking_pose_stamped_front);
+  }
+
+  next_undock_angle = angle + (abs(config.undock_angle) * (M_PI + M_PI) / 360.0) / 5.0;
+  if (next_undock_angle > abs(config.undock_angle) * (M_PI + M_PI) / 360.0)
+    next_undock_angle = -abs(config.undock_angle) * (M_PI + M_PI) / 360.0;
 
   exePathGoal.path = path;
   exePathGoal.angle_tolerance = 1.0 * (M_PI / 180.0);
--- a/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
@@ -55,9 +55,9 @@ Behavior *UndockingBehavior::execute() {
   docking_pose_stamped_front.pose = pose.pose.pose;
   docking_pose_stamped_front.header = pose.header;
 
-  int undock_point_count = 10;
-  double incremental_distance = config.undock_distance / undock_point_count;
-  for (int i = 0; i < undock_point_count; i++) {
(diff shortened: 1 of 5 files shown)


src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -47,16 +51,41 @@ Behavior *UndockingBehavior::execute() {
...
+    docking_pose_stamped_front.pose.position.x -= cos(yaw) * 0.1;
+    docking_pose_stamped_front.pose.position.y -= sin(yaw) * 0.1;
     path.poses.push_back(docking_pose_stamped_front);
   }
 
+  double angle;
+  if (config.undock_fixed_angle)
+    angle = config.undock_angle * (M_PI + M_PI) / 360.0;
+  else
+    angle = next_undock_angle;
```
rovo89, line 69: How about this?
```c++
double angle = config.undock_fixed_angle ? config.undock_angle * M_PI / 180.0 : next_undock_angle;
```

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -47,16 +51,41 @@ Behavior *UndockingBehavior::execute() {
...
+
+  undock_point_count = config.undock_angled_distance * 10.0;
+  for (int i = 0; i < undock_point_count; i++) {
+    double orientation;
+    if (config.undock_use_curve)
+      orientation = yaw + ((i + 1) * angle / undock_point_count);
+    else
+      orientation = yaw + angle;
```
rovo89, line 77: How about this?
```c++
for (int i = 1; i <= undock_point_count; i++) {
    double orientation = yaw + angle * (config.undock_use_curve ? i / undock_point_count : 1);
```

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -47,16 +51,41 @@ Behavior *UndockingBehavior::execute() {
...
+    docking_pose_stamped_front.pose.position.x -= cos(orientation) * 0.1;
+    docking_pose_stamped_front.pose.position.y -= sin(orientation) * 0.1;
+    tf2::Quaternion q(0.0, 0.0, orientation);
+    docking_pose_stamped_front.pose.orientation = tf2::toMsg(q);
+    path.poses.push_back(docking_pose_stamped_front);
+  }
+
+  next_undock_angle = angle + (abs(config.undock_angle) * (M_PI + M_PI) / 360.0) / 5.0;
+  if (next_undock_angle > abs(config.undock_angle) * (M_PI + M_PI) / 360.0)
+    next_undock_angle = -abs(config.undock_angle) * (M_PI + M_PI) / 360.0;
```
rovo89, line 87:
```c++
const double undock_angle_rad = abs(config.undock_angle) * M_PI / 180.0;
...
next_undock_angle = angle + undock_angle_rad / 5.0;
if (next_undock_angle > undock_angle_rad) {
    next_undock_angle = -undock_angle_rad;
}
```

rovo89:
I added some suggestions to make the code more crisp. I didn't check completely whether it makes sense, I think you tested that emperically. 😉

AndreKR:
Pete_MI632 had some changes which I have incorporated. I split them into separate commits to make the review easier: https://github.com/ClemensElflein/open_mower_ros/pull/151/commits/7acbeae35f1016073a28833195c5679e229a06e7 https://github.com/ClemensElflein/open_mower_ros/pull/151/commits/1ee783bef6bc860fac6b4fae0086e51d6e65a832 https://github.com/ClemensElflein/open_mower_ros/pull/151/commits/b951aa74627941b9e6ce6cae7563ae06f60c9420

I have then also applied your proposals as far as they were still valid.

I'm marking this as a draft though until I have empirically tested it again, it's raining currently.

image

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: With the random angle feature the `if` isn't as simple anymore, so I have not applied this change.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: I converted this into a ternary expression, but if you don't mind I'd like to keep the `i + 1` because I much prefer a 0-based loop instead of a 1-based loop.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: This code has been replaced with the random angle code (https://github.com/ClemensElflein/open_mower_ros/commit/1ee783bef6bc860fac6b4fae0086e51d6e65a832), so there's nothing to change anymore.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -37,7 +37,7 @@ std::string UndockingBehavior::state_name() {
 }
 
 Behavior *UndockingBehavior::execute() {
-  static double next_undock_angle = 0.0;
+  static bool seedRqd = true;
```
rovo89, line 40: The name should use under_scores instead of camelCase like the remaining code and "rqd" is unclear. Maybe `random_seeded = true` and then reverse the conditions?

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -55,37 +55,45 @@ Behavior *UndockingBehavior::execute() {
   docking_pose_stamped_front.pose = pose.pose.pose;
   docking_pose_stamped_front.header = pose.header;
 
-  int undock_point_count = config.undock_distance * 10.0;
+  int undock_point_count = 10;
```
rovo89, line 58: Giving the `10` a name for clarity is fine for me, but it should be `const` then.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -55,37 +55,45 @@ Behavior *UndockingBehavior::execute() {
...
+    ROS_INFO_STREAM("Fixed angle undock: " << config.undock_angle);
+  } else {
+    // seed based on first undock time rather than boot so should be ok even without RTC
+    if (seedRqd) {
+      srand(ros::Time::now().toSec());
+      ROS_INFO_STREAM("Random angle undock: Seeded rand()");
+      seedRqd = false;
+    }
+    double ranNum = (((((double)rand()) / RAND_MAX) - 0.5) * 2.0);
+    double ranAngle = abs(config.undock_angle) * ranNum;
```
rovo89, line 78: That looks overly complex for what it actually does (come up with a random angle between -undock_angle and +undock_angle). Also, naming conventions.
Suggestion:
```c++
double random_number = ((double) rand() / RAND_MAX) * 2.0 - 1.0;
double random_angle_deg = abs(config.undock_angle) * random_number;
```

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -55,37 +55,45 @@ Behavior *UndockingBehavior::execute() {
...
+    ROS_INFO_STREAM("Random angle undock: " << ranAngle);
+    angle = ranAngle * (M_PI + M_PI) / 360.0;
```
rovo89, line 80: `angle = random_angle_deg * M_PI / 180.0;`

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -55,37 +55,45 @@ Behavior *UndockingBehavior::execute() {
...
+  }
 
-  undock_point_count = config.undock_angled_distance * 10.0;
+  undock_point_count = 10;
```
rovo89, line 83: No need to set it again.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
```diff
@@ -55,37 +55,45 @@ Behavior *UndockingBehavior::execute() {
...
   for (int i = 0; i < undock_point_count; i++) {
-    double orientation;
-    if (config.undock_use_curve)
-      orientation = yaw + ((i + 1) * angle / undock_point_count);
-    else
-      orientation = yaw + angle;
-    docking_pose_stamped_front.pose.position.x -= cos(orientation) * 0.1;
-    docking_pose_stamped_front.pose.position.y -= sin(orientation) * 0.1;
-    tf2::Quaternion q(0.0, 0.0, orientation);
+    double orientation = yaw + angle * (config.undock_use_curve ? ((i + 1) / undock_point_count) : 1);
```
rovo89, line 86: I still feel that starting the loop with 1 would be easier, but I don't insist on it. Do overcome the integer division problem, `i + 1.0` should work.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: I guess that means "required", so how about `rng_seeding required`?

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: Done.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: Done.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: That's because it's actually two different "configuration" values that just happened to be both 10. I split them up.

src/mower_logic/src/mower_logic/behaviors/UndockingBehavior.cpp
AndreKR: Done.

AndreKR:
Where I work that's reserved for the reviewer.

rovo89:
I see. Well, my opinion is that I did review the code, even if I have no push permissions and nobody needs to follow my ideas. 😁  For me it's easier to see whether anything needs further discussion if the "done" or "won't do" boxes are minimized (they can still be expanded at any time).

AndreKR:
Real life test was successful as well. :shipit:

Apehaenger:
Cool PR! Like these undocking enhancement variations ;-)

Tested today without config settings as well as with config settings. What I didn't tested is `OM_UNDOCK_USE_CURVE=false` but that looks pretty plain to me.

However, beside @rovo89 's good comments, I think we need to change/add some tiny stuff before merging:
1. Need to add and describe the options to config/mower_config.* (also OM_DOCKING_APPROACH_DISTANCE)
2. The restored `OM_DOCKING_APPROACH_DISTANCE` need a default value so that it doesn't break mower configurations which don't have that env var
3. as well as config `docking_approach_distance` look like it has a wrong min value. min=0 looks wrong to me?! Shouldn't it be >0?

I know `docking_approach_distance` is not your's, but if we reintegrate the env var, it shouldn't pain someone ;-)

How are your thoughts?

AndreKR:
I have added the parameters and the defaults to the various schema and config files. Pete_MI632 had actually provided me with an example config.sh but I had no idea where to put it.

> config docking_approach_distance look like it has a wrong min value. min=0 looks wrong to me?! Shouldn't it be >0?

I don't think there is a logical issue with zero approach distance, it would just mean the robot drives to right in front of the dock with no chance to correct its heading. If your dock is at the end of a long narrow navigation area it might actually work.

Apehaenger:
Thanks a lot for the corrections!
Config samples got placed to the right place and the descriptions are quite good.

Regarding zero approach distance:
https://github.com/ClemensElflein/open_mower_ros/blob/c463bce2724739b9c0233ef5e64311088d3c55f3/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp#L45 Would become zero if approach distance is zero. Wouldn't it abort docking?
Unfortunately I can't test today ;-(

Apehaenger:
Openmower service exit with error 125 :-/
```RLException: $(env var) command only accepts one argument [env OM_DOCKING_APPROACH_DISTANCE 1.5]```

Fixed in local...

Tested with `export OM_DOCKING_APPROACH_DISTANCE=0` docking fails:
```
ct 11 09:58:33 Maehlanie openmower_launch[148104]: [ INFO] [1728633513.037735197]: --------------------------------------
Oct 11 09:58:33 Maehlanie openmower_launch[148104]: [ INFO] [1728633513.037779826]: - Entered state: DOCKING
Oct 11 09:58:33 Maehlanie openmower_launch[148104]: [ INFO] [1728633513.037816622]: --------------------------------------
Oct 11 09:58:33 Maehlanie openmower_launch[148103]: [ INFO] [1728633513.072849376]: Getting Docking Point
Oct 11 09:58:33 Maehlanie openmower_launch[148104]: [ WARN] [1728633513.073338187]: Waiting for good GPS
Oct 11 09:58:33 Maehlanie openmower_launch[148104]: [ WARN] [1728633513.166280943]: #### om_mower_logic: setMowerEnabled(0, 1) call
Oct 11 09:58:33 Maehlanie openmower_launch[148071]: [ INFO] [1728633513.188755583]: Setting mow enabled to 0
Oct 11 09:58:33 Maehlanie openmower_launch[148104]: [ INFO] [1728633513.189054561]: successfully set mower enabled to 0 (direction 1)
Oct 11 09:58:33 Maehlanie openmower_launch[148104]: [ WARN] [1728633513.189171986]: #### om_mower_logic: setMowerEnabled(0, 1) call completed within 0.0229266s
Oct 11 09:58:34 Maehlanie openmower_launch[148104]: [ INFO] [1728633514.073638245]: Calculating approach path
Oct 11 09:58:34 Maehlanie openmower_launch[148082]: [ERROR] [1728633514.096619510]: Failed to get a plan.
Oct 11 09:58:39 Maehlanie openmower_launch[148082]: [ INFO] [1728633519.079386291]: Global planner patience has been exceeded! Cancel planning...
Oct 11 09:58:39 Maehlanie openmower_launch[148082]: [ WARN] [1728633519.079631178]: Cancel planning failed or is not supported by the plugin. Wait until the current planning finished!
Oct 11 09:58:39 Maehlanie openmower_launch[148082]: [ INFO] [1728633519.084645357]: Planning patience (5s) has been exceeded; planner canceled!
Oct 11 09:58:39 Maehlanie openmower_launch[148082]: [ WARN] [1728633519.085019891]: No Recovery Behaviors loaded!
Oct 11 09:58:39 Maehlanie openmower_launch[148082]: [ WARN] [1728633519.085314666]: Abort the execution of the planner: Global planner exceeded the patience time
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ERROR] [1728633519.087107262]: Error during docking approach.
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ERROR] [1728633519.087230520]: Retrying docking approach
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ INFO] [1728633519.087295038]: 
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ INFO] [1728633519.087346297]: 
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ INFO] [1728633519.087392537]: --------------------------------------
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ INFO] [1728633519.087487740]: - Entered state: DOCKING
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ INFO] [1728633519.087542869]: --------------------------------------
Oct 11 09:58:39 Maehlanie openmower_launch[148103]: [ INFO] [1728633519.152488532]: Getting Docking Point
Oct 11 09:58:39 Maehlanie openmower_launch[148104]: [ WARN] [1728633519.153609096]: Waiting for good GPS
Oct 11 09:58:40 Maehlanie openmower_launch[148104]: [ INFO] [1728633520.153826672]: Calculating approach path
Oct 11 09:58:40 Maehlanie openmower_launch[148082]: [ERROR] [1728633520.172950674]: Failed to get a plan.
Oct 11 09:58:40 Maehlanie openmower_launch[148104]: [ INFO] [1728633520.666443852]: GPS quality: 0.95
Oct 11 09:58:42 Maehlanie openmower_launch[148078]: [ERROR] [1728633522.247679073]: [driver_gps] waited 13 ms to write to the tx buffer, serial port is probably congested!
Oct 11 09:58:45 Maehlanie openmower_launch[148082]: [ INFO] [1728633525.159207488]: Global planner patience has been exceeded! Cancel planning...
Oct 11 09:58:45 Maehlanie openmower_launch[148082]: [ WARN] [1728633525.159386598]: Cancel planning failed or is not supported by the plugin. Wait until the current planning finished!
Oct 11 09:58:45 Maehlanie openmower_launch[148082]: [ INFO] [1728633525.166662424]: Planning patience (5s) has been exceeded; planner canceled!
Oct 11 09:58:45 Maehlanie openmower_launch[148082]: [ WARN] [1728633525.167286993]: No Recovery Behaviors loaded!
Oct 11 09:58:45 Maehlanie openmower_launch[148082]: [ WARN] [1728633525.167431862]: Abort the execution of the planner: Global planner exceeded the patience time
Oct 11 09:58:45 Maehlanie openmower_launch[148104]: [ERROR] [1728633525.169145532]: Error during docking approach.
Oct 11 09:58:45 Maehlanie openmower_launch[148104]: [ERROR] [1728633525.169271642]: Retrying docking approach
Oct 11 09:58:45 Maehlanie openmower_launch[148104]: [ INFO] [1728633525.169335382]: 

...

Oct 11 09:59:03 Maehlanie openmower_launch[148104]: [ERROR] [1728633543.271717229]: Giving up on docking
```

AndreKR:
> Openmower service exit with error 125 :-/
> RLException: $(env var) command only accepts one argument [env OM_DOCKING_APPROACH_DISTANCE 1.5]

Oh, I see, the other values use `optenv` instead of `env`. Sorry, fixed.

> Tested with export OM_DOCKING_APPROACH_DISTANCE=0 docking fails:

Hm, yeah, so ideally we should have a different minimum there. I googled a bit and those parameters don't seem to support an exclusive min value, so we would have to come up with a minimum that isn't 0. Something like 1cm?

Apehaenger:
> > Openmower service exit with error 125 :-/
> > RLException: $(env var) command only accepts one argument [env OM_DOCKING_APPROACH_DISTANCE 1.5]
> 
> Oh, I see, the other values use `optenv` instead of `env`. Sorry, fixed.

Thanks!


> 
> > Tested with export OM_DOCKING_APPROACH_DISTANCE=0 docking fails:
> 
> Hm, yeah, so ideally we should have a different minimum there. I googled a bit and those parameters don't seem to support an exclusive min value, so we would have to come up with a minimum that isn't 0. Something like 1cm?

Well I think we should find a reasonable value. If I understand the functionality of "docking approach instance" in the right way, then it's the distance of the the mower to the dock when it tries to (re-)dock.
So I would think it should become some reasonable value like 1m, but I'm unsure what effect it has for more cramped docking locations :-/

Are you able to run a simulator?
If not, I would go i.e. with 0.5m even if it's probably unrealistic, most important is, that it doesn't abort again ;-)

AndreKR:
Yes, when you click "Record docking" the mower records its current GPS coordinates and heading. Then whenever it wants to go home, it will drive to a point that lays `OM_DOCKING_APPROACH_DISTANCE` meters before that recorded point in the recorded heading. At that point the mower itself will not yet have the correct heading in my experience. It will then navigate from that point to the recorded docking point, which gives it the correct heading. I guess that is in fact the intention behind the procedure. After that it will blindly drive forward `OM_DOCKING_DISTANCE` which will hopefully guide it into the rails of the dock.

I'm currently using an `OM_DOCKING_APPROACH_DISTANCE` of `1.0` and that is the absolute maximum of space I have available in front of the dock. I might have to reduce it a bit. (The actually required space is a bit more because it needs space to turn at the approach point.) For me `0.5` would work as a minimum.

I haven't gotten the simulator to work because it needs(?) rviz which in turn needs OpenGL and crashes when I try to run it over a forwarded X11 connection.

Apehaenger:
Thanks for the description.
In past we had to record two docking points. X1: 1-2m straight before the docking station and X2 nearly within the docking station. From the latter it drive then `OM_DOCKING_DISTANCE` till it get power.

So the question is: Does `OM_DOCKING_APPROACH_DISTANCE` is targeted to X1 or X2 or even skip X1?!

If `OM_DOCKING_APPROACH_DISTANCE` is targeting X1 case we could really use something like 0.01 as a realistic minimum.
But if it is targeting X2 then it should be something larger than the docking stations 1/2 width? Because with a smaller value it would already hit the dock and try to dock then?(theoretically spoken)

I also don't know how to run the simulator :-( 
However, testing could be easily be done by trying 0.01. My X1 is about 2m before the dock.
Please be so kind and test when you find some time, I'll also try, but at the moment its constantly raining here :-/

AndreKR:
> So the question is: Does OM_DOCKING_APPROACH_DISTANCE is targeted to X1 or X2 or even skip X1?!

Yes, I think `OM_DOCKING_APPROACH_DISTANCE` replaces X1. In other words, X1 is calculated from X2 (+ heading) and `OM_DOCKING_APPROACH_DISTANCE`.

> But if it is targeting X2 then it should be something larger than the docking stations 1/2 width? Because with a smaller value it would already hit the dock and try to dock then?(theoretically spoken)

Good point. In fact, I said above

> (The actually required space is a bit more because it needs space to turn at the approach point.)

without realizing this also restricts the minimum approach distance. So the approach distance should basically be about the length of the robot (it doesn't turn around its center).

Apehaenger:
Did 2 more tests today.
In the first I wanted to see if it targets X1 or X2, so I set a small `OM_DOCKING_APPROACH_DISTANCE=0.1` value. Surprisingly that also result in a false return of `approach_docking_point()` with the result of an aborted of docking approach!!

So, I thought it's a <1.0 thingy and tested with `OM_DOCKING_APPROACH_DISTANCE=0.95` which I also expected will fail, but that succeed!!

Its worthless to poke around this way :disappointed_relieved: . It most likely also depends on `docking_pose_stamped`.
We either ensure that https://github.com/AndreKR/open_mower_ros/blob/65382e8f24ef353c618c05545571aee10d9e41ca/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp#L32-L54 doesn't return false (docking aborts), or we remove `OM_DOCKING_APPROACH_DISTANCE` again from mower_config, so that no normal meat eating guy walk into this trap.

AndreKR:
😢 You don't know in what way the calculation fails when it's too small? Shall we put `1.0` as minimum then?

>  or we remove OM_DOCKING_APPROACH_DISTANCE again from mower_config

I really don't like having hidden config variables, the fact that it was missing from the example config felt very much like a bug. So before we remove it again, we should just put comments like "this can cause docking to fail if it's too small".

Apehaenger:
> 😢 You don't know in what way the calculation fails when it's too small? Shall we put `1.0` as minimum then?

I'm sorry "no" :cry:  Not clever enough :worried: . May be @rovo89 ?

> 
> > or we remove OM_DOCKING_APPROACH_DISTANCE again from mower_config
> 
> I really don't like having hidden config variables, the fact that it was missing from the example config felt very much like a bug. So before we remove it again, we should just put comments like "this can cause docking to fail if it's too small".

Not a good idea in my mind. It's forgotten 2 seconds after changed and saved.
BTW: That's exactly why hidden config variables exists. If they're not save, do **not** show them public and in special don't write "Hey I've a cool feature here, but it's buggy and may leave your mower somewhere sitting around" :scream:

Apehaenger:
> I didn't follow your conversation... In the highlighted lines, I think it's just about navigating to a certain point. I really lack knowledge in that the FTC Planner is doing and when it will give up. Could it be something simply like if the target point isn't within a navigation/mowing area and therefore not reachable?
> 
@AndreKR mainly search for a save minimum `OM_DOCKING_APPROACH_DISTANCE` value.
0.0 and 0.1 fail, 0.95 not, thus I said as long as we do not know what a save value is, we should not publish it in mower_config (but leave it hidden in dynamic reconf values).

I guess that the reasoning is somewhere here: https://github.com/AndreKR/open_mower_ros/blob/65382e8f24ef353c618c05545571aee10d9e41ca/src/mower_logic/src/mower_logic/behaviors/DockingBehavior.cpp#L45-L46 , but I unfortunately have been @ work while they talked in school about sin, cos, yaw, and the like :worried:

rovo89:
My understanding is that it simply calculates the target coordinates by starting from `docking_approach_point` and then moving `config.docking_approach_distance` meters away from it. `yaw` is the angle used for docking, hence subtracting to go into the opposite direction. Maybe [this](https://www.mathsisfun.com/geometry/unit-circle.html), especially [this image](https://www.mathsisfun.com/geometry/images/unit-circle-sin-cos-tan.svg) helps? `cos/sin(yaw)` really just calculate how far you have to go into x/y direction if you want to go in total one meter (radius 1) into the given direction. If you want to go two meters into that direction, you simply have to go twice the distance in x/y etc., therefore the multiplication.

So I doubt that the calculation is wrong, I think there's a problem navigating to the calculated point. Again, how far away is your docking station from the lawn? Have you covered it with a navigation area?

AndreKR:
Oooh, that would make sense... you probably recorded your mowing/navigation area by passing the dock, right? So a approach point far away from the dock would be within the navigation area, but an approach point too close would be outside of it.

Does it fail going from the lawn to the approach point or does it fail going from the approach point to the dock point?

Apehaenger:
> So I doubt that the calculation is wrong, I think there's a problem navigating to the calculated point. Again, how far away is your docking station from the lawn? Have you covered it with a navigation area?

No, I never covered any of my docks with a navigation area.
In my understanding the nav-area around the dock was always only a workaround for some specific cases. Also, I'm pretty sure that it was designed that way by Clemens that only dock-X1 need to be within a mow- or nav- area, whereas dock-x2 was planned to be possible outside any area (which is pretty clever!)
Beside that I can't cover any of my docking points because they're always outside the mow-area without space driving around. The one I tested yesterday is pretty simple and should always work in my opinion.

grafik


> 
> Does it fail going from the lawn to the approach point or does it fail going from the approach point to the dock point?

It fail immediately after pressing home (i.e. while mowing). Press Homes -> mower stops, tries to calculate whatever docking and abort docking.

rovo89:
I'm not too sure where in that screenshot the `docking_pose_stamped` is. I guess it's where the house icon is shown? And `docking_approach_point` would be `config.docking_approach_distance` in front of it, right? I always get confused which points are actually saved...

Anyway, I'm pretty sure that every point that you want to reach with the FTC Planner must be within the mowing/navigation areas. It will consider the cost map to drive around obstacles etc. With the default `config.docking_approach_distance`, you should always have a target point in the green area, but with a very small value, I can imagine that it's outside.

It would be easy to test this theory by recording a dummy docking station. The drive-straight-forward part won't succeed obviously. but it should navigate to it. Or just carry the mower while recording a navigation area 😆

rovo89:
Or even easier, just create the navigation area with Cedric's GUI by clicking on the points.

Apehaenger:
> I'm not too sure where in that screenshot the `docking_pose_stamped` is. I guess it's where the house icon is shown? And `docking_approach_point` would be `config.docking_approach_distance` in front of it, right? I always get confused which points are actually saved...
> 
Mee too :thinking: that's why I asked a couple of post before if `docking_approach_distance` is targeting dock-x1 or dock-x2.


> Anyway, I'm pretty sure that every point that you want to reach with the FTC Planner must be within the mowing/navigation areas. It will consider the cost map to drive around obstacles etc. With the default `config.docking_approach_distance`, you should always have a target point in the green area, but with a very small value, I can imagine that it's outside.
> 

That's the best fitting explanation till now and also would make sense to my failed as well working tries :+1: . `docking_approach_distance` is measured (or included in planner calc) from dock-x2 and thus need to be at least dock-x2 till hitting green(or white).
Simple to test :smile:

Apehaenger:
Checked the sources once more and @rovo89's description fit best. It is the home/dock position we've talked of.
So I did another test today. 
My dock is about 0.2m outside the mow-area. Tried with `OM_DOCKING_APPROACH_DISTANCE=0.3`.
For sure, not a very realistic value as one normally will use a much larger value. But it was pretty impressive to watch the mower how he drove directly before dock and curved into it :+1: 

I've been that insolent and changed 2 files directly in your branch, hope it's okay. Please validate.
Will do a last test in the evening or tomorrow if I didn't oversee a default value which I commented now in mower_config.
Probably you can test before?

Apehaenger:
Quite thanks for this PR :+1: . Just merged.

PR #153, opened 2024-09-14
Add OM_NO_COMMS env var for disabling comms node

jeremysalwen:
This is used by mowgli since it does not use the comms node.


ClemensElflein:
Why set the default so that it breaks for all the open mower users?

jeremysalwen:
Oops, forgot to change the default.

ClemensElflein:
Thank you for the update

PR #161, opened 2024-10-05
removed legacy relay

ClemensElflein:


ClemensElflein:
turns out the relay is very much in use for relaying the controller configurations.

PR #162, opened 2024-10-14
Add manual mower control actions in AreaRecording mode

olliewalsh:
Previously users would call mower_comms setMowEnabled directly to enable the mower while in AreaRecording mode.
However this should not have been possible since AreaRecording::mower_enabled() always returned false.

mower_logic now confirms that the mower state is as expected, so bypassing it in this way no longer works.

Add actions to toggle AreaRecording::mower_enabled() instead.

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.cpp
@@ -152,6 +152,7 @@ void AreaRecordingBehavior::enter() {
   has_outline = false;
   is_mowing_area = false;
   is_navigation_area = false;
+  manual_mowing = false;
 
   update_actions();
 
@@ -217,8 +218,7 @@ bool AreaRecordingBehavior::needs_gps() {
 }
 
 bool AreaRecordingBehavior::mower_enabled() {
-  // No mower during docking
-  return false;
+  return manual_mowing;
 }
 
 void AreaRecordingBehavior::pose_received(const xbot_msgs::AbsolutePose::ConstPtr &msg) {
@@ -578,6 +578,12 @@ void AreaRecordingBehavior::handle_action(std::string action) {
   } else if (action == "mower_logic:area_recording/collect_point") {
     ROS_INFO_STREAM("Got collect point");
     collect_point = true;
+  } else if (action == "mower_logic:area_recording/start_manual_mowing") {
+    ROS_INFO_STREAM("Starting manual mowing");
+    manual_mowing = true;
+  } else if (action == "mower_logic:area_recording/stop_manual_mowing") {
+    ROS_INFO_STREAM("Stopping manual mowing");
+    manual_mowing = false;
   }
   update_actions();
 }
@@ -633,6 +639,16 @@ AreaRecordingBehavior::AreaRecordingBehavior() {
   collect_point_action.enabled = false;
   collect_point_action.action_name = "Collect point";
 
+  xbot_msgs::ActionInfo start_manual_mowing_action;
+  start_manual_mowing_action.action_id = "start_manual_mowing";
+  start_manual_mowing_action.enabled = false;
+  start_manual_mowing_action.action_name = "Start manual mowing";
+
+  xbot_msgs::ActionInfo stop_manual_mowing_action;
+  stop_manual_mowing_action.action_id = "stop_manual_mowing";
+  stop_manual_mowing_action.enabled = false;
+  stop_manual_mowing_action.action_name = "Stop manual mowing";
+
   actions.clear();
   actions.push_back(start_recording_action);
   actions.push_back(stop_recording_action);
@@ -644,6 +660,8 @@ AreaRecordingBehavior::AreaRecordingBehavior() {
   actions.push_back(auto_point_collecting_enable_action);
   actions.push_back(auto_point_collecting_disable_action);
   actions.push_back(collect_point_action);
+  actions.push_back(start_manual_mowing_action);
+  actions.push_back(stop_manual_mowing_action);
 }
 
 void AreaRecordingBehavior::update_actions() {
@@ -681,6 +699,10 @@ void AreaRecordingBehavior::update_actions() {
         actions[6].enabled = true;
       }
     }
+    // start_manual_mowing
+    actions[10].enabled = !manual_mowing;
+    // stop manual mowing
+    actions[11].enabled = manual_mowing;
 
     registerActions("mower_logic:area_recording", actions);
   }
--- a/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.h
+++ b/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.h
@@ -89,6 +89,8 @@ class AreaRecordingBehavior : public Behavior {
   bool auto_point_collecting = true;
   bool collect_point = false;
 
+  bool manual_mowing = false;
+
   visualization_msgs::MarkerArray markers;
   visualization_msgs::Marker marker;
 
--- a/src/mower_logic/src/mower_logic/mower_logic.cpp
+++ b/src/mower_logic/src/mower_logic/mower_logic.cpp
@@ -75,6 +75,7 @@ ros::Time pose_time(0.0);
 xbot_msgs::AbsolutePose last_pose;
 ros::Time status_time(0.0);
 mower_msgs::Status last_status;
+ros::Time joy_vel_time(0.0);
 
 ros::Time last_good_gps(0.0);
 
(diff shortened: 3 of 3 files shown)


Apehaenger:
Quite thanks for the PR!
`mow_enabled` was a quite useful service while developing as well during assembly.

Just fetched and compiled, but unfortunately does now work as I expected :cry: (removed GPS messages)
```
Oct 15 12:43:27 Maehlanie openmower_launch[995]: [ INFO] [1728989007.135884492]: --------------------------------------
Oct 15 12:43:27 Maehlanie openmower_launch[995]: [ INFO] [1728989007.136100108]: - Entered state: AREA_RECORDING
Oct 15 12:43:27 Maehlanie openmower_launch[995]: [ INFO] [1728989007.136308020]: --------------------------------------
Oct 15 12:43:27 Maehlanie openmower_launch[1006]: [ INFO] [1728989007.150932327]: new actions registered: mower_logic:area_recording registered 12 actions.
Oct 15 12:43:27 Maehlanie openmower_launch[995]: [ INFO] [1728989007.154198177]: successfully registered actions for mower_logic:area_recording
Oct 15 12:43:27 Maehlanie openmower_launch[995]: [ INFO] [1728989007.162463657]: Starting recording area
Oct 15 12:43:27 Maehlanie openmower_launch[995]: [ INFO] [1728989007.162636754]: Subscribing to /joy for user input
Oct 15 12:43:34 Maehlanie openmower_launch[949]: [ INFO] [1728989014.247301264]: Setting mow enabled to -1
Oct 15 12:43:34 Maehlanie openmower_launch[995]: [ WARN] [1728989014.412352229]: #### om_mower_logic: setMowerEnabled(0, 0) call
Oct 15 12:43:34 Maehlanie openmower_launch[949]: [ INFO] [1728989014.420841717]: Setting mow enabled to 0
Oct 15 12:43:34 Maehlanie openmower_launch[995]: [ INFO] [1728989014.421100944]: successfully set mower enabled to 0 (direction 0)
Oct 15 12:43:34 Maehlanie openmower_launch[995]: [ WARN] [1728989014.421219798]: #### om_mower_logic: setMowerEnabled(0, 0) call completed within 0.00889661s
Oct 15 12:44:00 Maehlanie openmower_launch[949]: [ INFO] [1728989040.887257250]: Setting mow enabled to 1
Oct 15 12:44:00 Maehlanie openmower_launch[995]: [ WARN] [1728989040.912349533]: #### om_mower_logic: setMowerEnabled(0, 0) call
Oct 15 12:44:00 Maehlanie openmower_launch[949]: [ INFO] [1728989040.920927709]: Setting mow enabled to 0
Oct 15 12:44:00 Maehlanie openmower_launch[995]: [ INFO] [1728989040.921320341]: successfully set mower enabled to 0 (direction 0)
Oct 15 12:44:00 Maehlanie openmower_launch[995]: [ WARN] [1728989040.921570231]: #### om_mower_logic: setMowerEnabled(0, 0) call completed within 0.00920594s
Oct 15 12:44:24 Maehlanie openmower_launch[949]: [ INFO] [1728989064.537098896]: Setting mow enabled to 0
```

olliewalsh:
> Quite thanks for the PR! `mow_enabled` was a quite useful service while developing as well during assembly.
> 
> Just fetched and compiled, but unfortunately does now work as I expected 😢 (removed GPS messages)
> 
 
Are you expecting calling the `mower_service/mow_enabled` service to work? It will, but mower_logic will call it right after you do and turn it off again. Need to call the action from this patch to tell mower_logic to enable mowing instead e.g:
`rostopic pub -1 /xbot/action std_msgs/String mower_logic:area_recording/start_manual_mowing`

Apehaenger:
> 
> Are you expecting calling the `mower_service/mow_enabled` service to work? It will, but mower_logic will call it right after you do and turn it off again. 

Yip, had expected it :innocent: 

>
>Need to call the action from this patch to tell mower_logic to enable mowing instead e.g: `rostopic pub -1 /xbot/action std_msgs/String mower_logic:area_recording/start_manual_mowing`

Quite thanks for the command, did not know that before :shushing_face: and now after checking the sources, I do see.
However, unfortunately does not work as (I) expected.
Mow motor start sometimes for a couple of seconds and then stops again. If it doesn't start on first try, I also wasn't able to start it even with multiple tries.

Is it expected to call the command with a rate i.e. `rostopic pub -r 10 ...`?

Now, the danger part: When failed to start, exited area-recording and went back to area-recording. That immediately started the mow motor again. This is somehow dangerous when used during assembling :scream:  
So better set `manual_mowing = false;` when leaving area-recording?

olliewalsh:
> Mow motor start sometimes for a couple of seconds and then stops again. If it doesn't start on first try, I also wasn't able to start it even with multiple tries.

Safety checks in mower_logic still apply, so if GPS drops I expect it will stop the mower, which might explain this - https://github.com/ClemensElflein/open_mower_ros/blob/main/src/mower_logic/src/mower_logic/mower_logic.cpp#L491

Also may need to keep the mower moving or mower_comms check will stop the mower https://github.com/ClemensElflein/open_mower_ros/blob/main/src/mower_comms/src/mower_comms.cpp#L114

If we want to allow mowing with bad/no GPS might need to create a new behavior class instead of using AreaRecording

olliewalsh:
> So better set `manual_mowing = false;` when leaving area-recording?

Done, when entering area-recording

olliewalsh:
> Is it expected to call the command with a rate i.e. rostopic pub -r 10 ...?
No, just need to send it once to call the action


> Mow motor start sometimes for a couple of seconds and then stops again. If it doesn't start on first try, I also wasn't able to start it even with multiple tries.

Works fine in the sim. I'll need dust off a mower to try to reproduce this. mower_comms does stop the blade if no cmd_vel message has been received for 25s but IIUC the app should send a cmd_vel every 10s. @Apehaenger Can you confirm (do you recall) that the app was open at the time?

olliewalsh:
Setting PR to draft until I test with with a real mower

olliewalsh:
Tested with real mower today, works well

ClemensElflein:
Should we create a "Manual Mowing" Behavior for this instead?

patrickzzz:
"Mowing" in "area recording" feels wrong for me as well (while manual mowing is a cool feature).

Apehaenger:
> Should we create a "Manual Mowing" Behavior for this instead?

Good idea but because of @patrickzzz comment we probably might need to clarify/define some phrases/meaning.

I'm (personally) not interested in any possibility to do "manual mowing" like steering around the lawn and mow difficult areas by hand (even if it would be possible and other might like or need it).
I'm mainly interested in some kind of "manual mow test" to check: "does it drive straight, is it slow/fast enough, does it stop well, ..." and also **very** important: "does the blade-motor" work correctly, steady, measure RPM, ...

In past we did that simply in "area recording mode" because we had these possibilities there with the old `mow_enabled?` service.

If I understand @ClemensElflein comment right
> Should we create a "Manual Mowing" Behavior for this instead?

he doesn't mean to add a new "Manual Mowing" functionality, but unscramble/cleanup "AreaRecording" behavior from these kind of gimmicks, and better make a more clean and straight i.e. "Manual Mowing" behavior.

I'm still not familiar enough with ROS and also do **not** know the effort, but "yes" that sound like the best approach (if I understood him right).

However, I'm fine with both and would be happy if I don't need to patch ROS anymore to do some mow-motor tests.
So... in any way: Quite thanks for your PR @olliewalsh !!

Apehaenger (approved):
I'm still not really competent to approve ROS stuff, because it still has some mystic to me :smiling_face_with_tear: ...
... however looks good an logic to me.

In addition I don't see any risk that it could break major functionality.

Apehaenger:
> Yeah, did mention earlier in the comments that perhaps this should be another behavior.

Yes I remember that it got mentioned...

> I expect that will need support in the app too though.

Not in my opinion :innocent: 
If we can call/trigger it by some kind of ROS CLI call, I (personally) would be fine with it, and if someone wants a GUI button for it, he/she might add it :wink: 

> Initially just wanted to quickly restore the ability to turn on the mower in area recording as a number of people seem to have been using that.

Yes I also often used that. But as said, I'm fine with both.

You wrote that you tested it? With a real mower? No further testing required?

ClemensElflein:
OK, if we don't include it in the GUI, this is fine for me as well. If we want a real "drive the mower to manually mow an area"-feature, we should move it into a separate behavior later.

olliewalsh:
> Not in my opinion 😇
> If we can call/trigger it by some kind of ROS CLI call, I (personally) would be fine with it, and if someone wants a GUI button for it, he/she might add it 😉

I meant the app would need to know about the new manual mowing behaviour to show the virtual joystick etc...

PR #163, opened 2024-10-14
Enlargable Config Packet

Apehaenger:
In relation to https://github.com/ClemensElflein/OpenMower/pull/110

Roadmap:

- [x] Decide for config protocol and struct
- [x] LowLevel (OpenMower Pico-FW) please see https://github.com/ClemensElflein/OpenMower/pull/110
- [x] HighLevel (open_mower_ros)
  - [x] Implement packet handling
  - [x] Implement config value handling
    - [x] Charge & Battery
    - [x] Hall sensor inputs
    - [x] Emergency
    - [x] Misc
- [x] Test with old LL-FW

<sub>:hammer: = Currently working on</sub>

### DIFF
--- a/config/mower_config.schema.json
+++ b/config/mower_config.schema.json
@@ -506,6 +506,24 @@
                 "type": "number",
                 "description": "Use perimeter sensor for docking. Set to 1 or 2 depending on the signal selected on the docking station. This is only supported by Mowgli builds for now.",
                 "x-environment-variable": "OM_PERIMETER_SIGNAL"
+              },
+              "OM_EMERGENCY_INPUT_CONFIG": {
+                "type": "string",
+                "default": "",
+                "description": "Comma separated list of emergency inputs (halls or stop), where each sensor can be configured in one of the following modes '[!]<I>gnore|<U>nused|<S>top|<L>Lift'",
+                "x-environment-variable": "OM_EMERGENCY_INPUT_CONFIG"
+              },
+              "OM_EMERGENCY_LIFT_PERIOD": {
+                "type": "number",
+                "default": -1,
+                "description": "Period (ms) for >=2 wheels to be lifted in order to count as emergency (0 = disable)",
+                "x-environment-variable": "OM_EMERGENCY_LIFT_PERIOD"
+              },
+              "OM_EMERGENCY_TILT_PERIOD": {
+                "type": "number",
+                "default": -1,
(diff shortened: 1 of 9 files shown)


Apehaenger:
Intermediate commit #2 is only for illustration purposes

## What's in

- Flexible length config packet is straight forward and works
- Config Packet handling HL/LL, LL/mower_logic is illustrated by two LL specific x_charge_* config values

## What has changed

In past the (old) config packet request/response phase got trigger by LL whenever ROS became live (detected via heartbeat packet).
This got reverted now because it would result in at least one waste packet.
Read on how it's now...

## General config packet handling

Whenever HL detects missing LL-**status** packets for > 1 sec. it assumes that the mower just booted or LL rebooted i.e. due to flashing a new FW, and send a config packet (and response request) with HL's configured LL values, which are normally not any!

Yes, (in my opinion) HL does not need to know any specific LL values in >90?%, because it is HL.
LL knows about it's default C500 configuration values.
Except in the rare cases where someone is using a custom mower, battery, hall-config or whatever.
In those case, the individual LL values got placed in mower_config (not yet implemented. Only dynamic reconfigure parameters got added for illustration and tests) and get send within the initial config request packet.

LL will receive this HL-config packet, do a sanity check where it has hardware limits like (charge cutoff voltage/current), and send the used/adjusted valued back to HL as "response".

When HL receives LL's response, dependent on what/if something changed in comparison to what requested before, or did not know before (because HL does not know much about LL), it will send a dynamic reconfigure and inform mower_logic about the relevant/changed values, which are used i.e. for battery-low calculations (go home), sensor view, ...

Similar will happen if one is changing values via dynamic reconfigure (mower_logic). Change get compared with what is in, if changed will inform LL with a new config (including the response request) and will align once response get received.

Did a couple of tests with the two implemented vales. Works pretty good. A full packet round, even if with a to large value, take < 1s.

## What's bad

I don't like the static member checks in `manageDynReconfConfigValues()` and `manageLowLevelConfigValues()`.
Would like to iterate over the config struct members, but for this it looks like boost is required. See `FIXME` comment within these function. Any tips how it could be made more dynamic are welcome! :innocent:

Apehaenger:
Same here. In my opinion, all is done.
Did a couple of tests with old-LL/new-HL, new-LL/old-HL, new-LL/new-HL on C500 as well as SA600? mower.
Haven't found any issue.

Would love to see a generic review before asking for tester @ discord.

ClemensElflein (approved):
LGTM

PR #152, opened 2024-09-14
Actions update and small fixes

docgalaxyblock:

PR #165, opened 2024-10-17
Chore/fix ci

ClemensElflein:


ClemensElflein:
Requesting review from @docgalaxyblock and @Apehaenger because you both worked on the CI issue. @docgalaxyblock says its a good idea to keep the builder on `-latest`, but the clang-format failed due to version mismatch, this seems to fix it.

Apehaenger (approved):
I would also prefer to go with -latest instead of getting antique by time.
In fact I also tried to examine how, but all I found, sound wired (like set to `python@3` instead of `python@3.x.x`), so I ugly switched to ubuntu-22.04 because it was obvious that the new -latest (24.x) was the reason for the failing CI.

Specifying the clang-format version is also preferable, even though I couldn't identify what really happen in the discussed issue at Discord. But similar already happen before https://github.com/ClemensElflein/open_mower_ros/pull/141#discussion_r1715891778

I'm unsure about if you didn't already tested switching the auth username to `github.actor` during last try, but I even don't know how to test without merging :innocent:

docgalaxyblock:
OK give it a short check and my commits are already included in #152 where the `github.actor` did not worked and I therefore [force-pushed the commit out again](https://github.com/ClemensElflein/open_mower_ros/compare/4ea5efffeed2e4cdc96019f03cf9eaf50ec72798..b0641138de043711b518537722c26a30e254c0d4).

I would remove all commits except 28e4b383b86d317898ff4f96d8938efc15c66072 and then add my discovered fixes for the 24.04/latest based runner.

ClemensElflein:
force pushed your linked commits to this pr

docgalaxyblock (approved):
Looks good, the Workflow also succeeded.

PR #168, opened 2025-01-18
Use native arm64 github action runner

olliewalsh:
https://github.blog/changelog/2024-09-03-github-actions-arm64-linux-and-windows-runners-are-now-generally-available/

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -31,10 +31,15 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
-        docker_build_arch: [ linux/arm64 , linux/amd64 ]
+        os: [ ubuntu-24.04-arm, ubuntu-24.04 ]
+        include:
+          - os: ubuntu-24.04-arm
+            docker_build_arch: arm64
+          - os: ubuntu-24.04
+            docker_build_arch: amd64
     steps:
       - name: Prepare
         env:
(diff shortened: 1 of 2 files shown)


rovo89:
Available for free plans since January 16: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

ClemensElflein:
Cool! I didn't know they made them free now. Thank you for the PR. I have restarted the build, let's see if it will pass 🤞 The permission issue still persists though (and I still have no idea why)

ClemensElflein:
Hmm restarting the action also didn't let it run properly. I'll merge and see if that works

docgalaxyblock:
This change is totally useless. @olliewalsh Only renamed the matrix job names to some somefancy ones.
The wrkflow still uses amd64 and you removed the qemu setup step so a clean build should fil at the arm64 matrix run.

Lets's see if I assumeit correctly: https://github.com/ClemensElflein/open_mower_ros/actions/runs/12852509134

Also they are just a preview: https://github.com/orgs/community/discussions/148648

docgalaxyblock:
Im am right - the runner is amd64: https://github.com/ClemensElflein/open_mower_ros/actions/runs/12852509134/job/35834579604#step:9:133

I guess it did not creahed sinc qemu is now installed as default on gthub hosted runners.

olliewalsh:
> I guess it did not crashed since qemu is now included by default on github hosted runners.

ah, the matrix example makes it look like a different "os" can be used for each matrix job, but it's just an arbitrary label on the job https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy

ClemensElflein:
Thanks @docgalaxyblock for checking, I have reverted the PR

ClemensElflein:
Sorry for the premature merge - I thought it's just the permission issue

olliewalsh:
Will follow up with another PR, I thought that was taking too long :-) Initially tried building on the arm runner so it does work - just need to restructure the jobs

docgalaxyblock:
I am for reverting and then wait for the arm64 runners to no longer be in preview.
After that we can develop a new multi run workflow since afaiu it is a bit moe complex to implement.

olliewalsh:
Just needs `runs-on: ${{ matrix.os }}` AFAICT

olliewalsh:
> I am for reverting and then wait for the arm64 runners to no longer be in preview.

Can we use switch for now and revert if we hit any issues? It's sooo much nicer having fast arm64 builds

ClemensElflein:
> > I am for reverting and then wait for the arm64 runners to no longer be in preview.
> 
> Can we use switch for now and revert if we hit any issues? It's sooo much nicer having fast arm64 builds

yes

docgalaxyblock:
As I said I am not for jumping on the hype train.
But you can add it to you forks main branch then the arm64 runner will be used for your other branches too.

Also if you can trigger a manual clean build after that we can see if the run runs without any problems

ClemensElflein:
There's really no risk involved in merging this

olliewalsh:
Risk I would think is that the queues could get long if there are not enough runners to keep up with demand

PR #175, opened 2025-03-28
Handle uart mapping changing on newer kernel

olliewalsh:
Since 6.1.28 the uart mappings have changed:
https://forums.raspberrypi.com/viewtopic.php?t=347868#p2106791

### DIFF
--- a/src/open_mower/launch/include/_comms.launch
+++ b/src/open_mower/launch/include/_comms.launch
@@ -9,6 +9,9 @@
         <rosparam unless="$(eval env('OM_MOWER')=='CUSTOM')" file="$(find open_mower)/params/hardware_specific/$(env OM_MOWER)/comms_general_params.yaml"/>
         <rosparam unless="$(eval env('OM_MOWER')=='CUSTOM')" file="$(find open_mower)/params/hardware_specific/$(env OM_MOWER)/comms_$(env OM_MOWER_ESC_TYPE)_params.yaml"/>
         <rosparam if="$(eval env('OM_MOWER')=='CUSTOM')" file="$(env HOME)/mower_params/comms_params.yaml"/>
+        <param if="$(eval optenv('OM_XESC_LEFT_PORT')!='')" name="left_xesc/serial_port" value="$(optenv OM_XESC_LEFT_PORT)"/>
+        <param if="$(eval optenv('OM_XESC_RIGHT_PORT')!='')" name="right_xesc/serial_port" value="$(optenv OM_XESC_RIGHT_PORT)"/>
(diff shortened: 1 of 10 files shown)


Apehaenger (commented):
Quite thanks for your trust that I could judge :innocent: 

I do like the dynamic approach in default_environment.sh but I'm unknown to the general workflow how those settings get merged together. So let me try to describe in words how I understand the code.

1. All related uarts in the hardware_specific yaml files get set to the >= 6.1.28 uarts
2. `default_environment.sh` will set `OM_XESC_X_PORT` to the older uart names if an older kernel get detected but only if `OM_XESC_X_PORT` isn't set (i.e. in mower_config)
3. Within the launch file, if the new `OM_XESC_X_PORT` is set, it will overwrite the serial port setting from the yaml file.

So, only those who manually adapted their serial port in their yaml definition, or those who have a separate yaml definition for their mower model, need to touch their configuration, preferable via an `OM_XESC_X_PORT` in their mower_config.

If I understood that right, than it looks clever to me :innocent: !

For sure this also happen to GPS but for ease of writing I mentioned only the ESC's.

BTW: Your PR also contains slightly increased x_motor_rpm values. I'm fine with them, but unsure of their reasoning.

olliewalsh:
> BTW: Your PR also contains slightly increased x_motor_rpm values. I'm fine with them, but unsure of their reasoning.

ah yea, meant to ask you about that. I changed it to the same as C500. Not sure why they were different for SA650 as it's the same motor.

Apehaenger:
Oh, not for my SA6x ... probably because it's a Rev4 SA. However, doesn't matter in real as those settings get only used for the gauge sensor display in GUI.

Apehaenger:
> Nope the core + chassis yfs have planetary gearbox motors and the c500 has a classic gearbox in the chassis

Hope you missed the middle part :laughing: ... because we're talking about the blade-motor :innocent: ... otherwise I've a big knowledge gap when there're blade-motors with a gearbox in the wild :innocent:

docgalaxyblock:
Wopsi 😇

ClemensElflein (approved):
Thank you for the PR! Looks good to me, any reason why this is still draft?

olliewalsh:
> Thank you for the PR! Looks good to me, any reason why this is still draft?

Just have not tested it on raspbian bullseye yet... Trying it quickly now...

olliewalsh:
> > Thank you for the PR! Looks good to me, any reason why this is still draft?
> 
> Just have not tested it on raspbian bullseye yet... Trying it quickly now...

Forgot to update - should be good to merge now

ClemensElflein:
nice! thank you!

PR #176, opened 2025-04-08
IMU axis mapping for V2

Apehaenger:
This is how I would implement a flexible configurable IMU axis mapping.
Not yet tested, also SABO axis sign might be wrong because untested.

How it's implemented:

yaml config file get a new entry:
```
mower_comms_v2:
    imu:
      axis_config: "+Z-X-Y"
```

Syntax of axis_config is <axis sign +|-><physical IMU axis which shall be used for ROS axis X. One of X|Y|Z>...(same for ROS axis Y and Z)

ImuServiceInterface::OnConfigurationRequested() will parse the string into two new ImuService Registers:
```
"registers": [
    {
      "id": 0,
      "name": "AxisRemap",
      "type": "uint8_t[3]"
    },
    {
      "id": 1,
      "name": "AxisSign",
      "type": "int8_t[3]"
    }
  ]
```  

which in turn get directly used in firmware during assignment of Axes[].

Also related:
[FW PR](https://github.com/xtech/fw-openmower-v2/pull/8)
[IMU service registers PR](https://github.com/xtech/definitions-open-mower/pull/1)

### DIFF
--- a/src/mower_comms_v2/src/ImuServiceInterface.cpp
+++ b/src/mower_comms_v2/src/ImuServiceInterface.cpp
@@ -19,3 +19,50 @@ void ImuServiceInterface::OnAxesChanged(const double* new_value, uint32_t length
   imu_msg.angular_velocity.z = new_value[5];
   imu_publisher_.publish(imu_msg);
 }
+
+bool ImuServiceInterface::validateAxisConfig() {
+  if (axis_config_.size() != 6) {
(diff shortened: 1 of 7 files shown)


Apehaenger:
Also added my Sabo params to this PR because it already had one integrated as IMU mapping example

ClemensElflein:
Looks good to me, thank you for the PR!

PR #179, opened 2025-04-16
Pause semi-auto mowing when rain detected and rain mode is DOCK

### LINKED ISSUE #178: rain detection - wrong/unexpected bahvior
Platform
-  Yardforce Classic 500
- OM_HARDWARE_VERSION="0_11_X_WT901" (Vermut's kit)
- recent lowlevel firmware (I hope)
- please note: openmower image recent _v2_ (2590be9) with _v1 configuration_ (don't know if relevant)

Relevant Configuration
```
  OM_AUTOMATIC_MODE=1 // SemiAuto
  OM_RAIN_MODE=1      // Dock
```

Observed Behavior
- mower is in state `mowing`
- I pour water over the rain sensor and trigger docking and mower docks => expected
- but undocks by itself when battery is full => not expected
- it goes to the undocking point and 
    - either immediately docks again probably because the sensor still yields `rain` it does it in a loop "dock -> battery full -> undock -> rain detected -> dock"
    - or it stays in `idle` mode at the undocking position (I don't know why)

Expected Behavior
With `OM_RAIN_MODE=1` I'd expect that the behavior is the same as if I had pressed the HOME-button, i.e. the mower should go to the dock and stay there even if battery is full and even the rain sensor yields `no_rain` later. The mower should only start mowing again on user action.

I did not understand the source code (or did not find the right position) but the behavior is as if the `OM_AUTOMATIC_MODE=1` overrides `OM_RAIN_MODE=1`

olliewalsh:
Fixes #178


Max07B2:
Sorry, here's probably the better place for comments on the solution (not the original issiue), see: https://github.com/ClemensElflein/open_mower_ros/issues/178#issuecomment-2811883089

PR #193, opened 2025-06-10
Switch license to GPLv3

jeremysalwen:
Switches the license to GPLv3

As discussed in #136 we have the permission of all nontrivial contributors to re-license as GPLv3.

A couple of notes:
- The license is GPLv3, not GPLv3-or-later, as contributors only approved GPLv3.
- I changed the copyright notice from "Clemens Elflein" to "Clemens Elflein and OpenMower contributors".
- I removed the section about not using the project to profit, and just put a bare-bones mention of the GPL license.  I wasn't sure if you wanted to put more text about how it's intended to be used, or just mentioning the GPL is clear enough.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,437 +1,674 @@
-Attribution-NonCommercial-ShareAlike 4.0 International
-
-=======================================================================
-
-Creative Commons Corporation ("Creative Commons") is not a law firm and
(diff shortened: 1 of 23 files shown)


ClemensElflein:
Thank you so much!

PR #205, opened 2025-06-19
Fix OM_NO_COMMS to only disable COMMS, not GPS

jeremysalwen:
For Mowgli, being able to disable comms separately from GPS is important. PR #174 changed this and replaced the separate OM_NO_COMMS and OM_NO_GPS with just OM_NO_COMMS that disabled *both*.  This broke mowgli.  This PR restores the original behavior of OM_NO_COMMS and OM_NO_GPS for v1.

I believe for V2 the comms and GPS are integrated so this will have no effect.


ClemensElflein:
LGTM, this is indeed a bug in the launch file, thank you for the PR

PR #211, opened 2025-06-24
Remove erroneous subst_value from _record.launch.

jeremysalwen:
This was causing a warning on startup.


rovo89:
Ah, was that from the other syntax you used first?

jeremysalwen:
Yes, exactly.  Not sure how it got in there.

PR #239, opened 2025-09-15
Rework Environment Variables

ClemensElflein:
Here are the changes:
- Instead of OM_V2 we now have HARDWARE_PLATFORM=2
- Instead of OM_MOWER_ESC_TYPE we now have ESC_TYPE

The goal is to have all current installs (loading config from environment) to not be affected

## Summary by CodeRabbit

- New Features
  - Added HARDWARE_PLATFORM to choose v1/v2 hardware behavior at runtime.
  - Introduced OM_LEGACY_CONFIG_MODE to enable legacy env-based parameter loading.
  - Exposes ESC_TYPE and MOWER in the runtime environment.

- Refactor
  - Renamed mower ESC config/env mapping from OM_MOWER_ESC_TYPE to ESC_TYPE; parameter selection and comms/params gating now use HARDWARE_PLATFORM and OM_LEGACY_CONFIG_MODE; NTRIP unchanged.

- Chore
  - Runtime user added to dialout group in the container image.

### DIFF
--- a/docker/openmower_entrypoint.sh
+++ b/docker/openmower_entrypoint.sh
@@ -13,8 +13,8 @@ fi
 
 if [[ "$MODE" == "osv1" ]]; then
     source /config/mower_config.sh
-    # If OM_V2 is set (to anything), set HARDWARE_PLATFORM=2 and new (yaml-based) config, else 1 and environment config
-    if [[ -n ${OM_V2+x} ]]; then
+    # If OM_V2 is truthy, set HARDWARE_PLATFORM=2 and new (yaml-based) config, else 1 and environment config
+    if [[ "${OM_V2,,}" =~ ^(true|1|yes)$ ]]; then
         export HARDWARE_PLATFORM=2
     else
         export HARDWARE_PLATFORM=1


docker/openmower_entrypoint.sh
```diff
@@ -13,10 +13,19 @@ fi
 
 if [[ "$MODE" == "osv1" ]]; then
     source /config/mower_config.sh
+    # If OM_V2 is set (to anything), set HARDWARE_PLATFORM=2, else 1
+    if [[ -n ${OM_V2+x} ]]; then
```
Apehaenger, line 17: Testing OM_V2 to anything looks dangerous to me :innocent:

I would go with something like: `[[ "${OM_V2,,}" =~ ^(true|1|yes)$ ]]`

Apehaenger (approved):
Looks okay to me, but I'm not that good with the launch files :innocent:

docker/openmower_entrypoint.sh
ClemensElflein: fixed in #05d123b

PR #240, opened 2025-09-16
Rework Environment

ClemensElflein:
Using the V1 hardware on V2 OS wasn't working properly because in OSv2 the default parameters are loaded through yaml instead of the environment.

Changes here:
- Adapted the _params.launch file to set the parameters similar to the yaml structure (but read from the same environment variables, so nothing should break)
- Adapted mower_comms_v1 to read from the yaml strucutre (which should now work for OSv2)

## Summary by CodeRabbit

- Refactor
  - Renamed the v1 comms node to "ll" and reorganized parameters into ll/services/... namespaces (diff_drive, sound), preserving runtime behavior.

- Chores
  - OSv1 environment handling adjusted: PARAMS_PATH and RECORDINGS_PATH are now exported and used.
  - Rosbag recordings now write to RECORDINGS_PATH instead of HOME.
  - Hardware-specific in-script default fallbacks removed in favor of YAML-driven defaults; OM_V2 flag set for affected hardware.


ClemensElflein:
Someone please test this with OSv1 / Hardware V1

PR #243, opened 2025-09-25
Fix container shell prefix and stick to color prompt

Apehaenger:
Moved container shell prefix to here instead of OpenMowerOS. See https://github.com/ClemensElflein/OpenMowerOS/pull/27

Draft, because not yet tested.


ClemensElflein:
@Apehaenger looks good to me, not sure how to resolve the conflicts right now. If its no issue for you, please do, otherwise I'll need to enable my brain at some point and will fix it.

PR #269, opened 2026-03-11
Provide explicit NTRIP version when connecting

ClemensElflein:
## Summary by CodeRabbit

* **Chores**
  * Updated the default NTRIP version to **"Ntrip/2.0"** in the NTRIP client launch configuration.

PR #273, opened 2026-03-27
Replace ntrip client with ros_ntrip_client

olliewalsh:
Drop the troublesome ntrip client and use https://github.com/olliewalsh/ros_ntrip_client instead.
It's an AI generated C++ ntrip client that focuses on complying with the recommended throttling logic for public provides to avoid getting the IP banned.

Confirmed working with rtk2go, centipede, sapos, and local rtkbase.

## Summary by CodeRabbit

* **New Features**
  * Upgraded NTRIP client with adaptive reconnection strategy featuring exponential backoff
  * Added configuration options for RTCM timeout and adaptive connection retry behavior


olliewalsh:
ghcr.io/olliewalsh/open_mower_ros:releases-v2_ntrip if anyone wants to test this

olliewalsh:
@coderabbitai review

Apehaenger (commented):
I'm not lighten enough with NTRIP and ROS to do a valuable review 😇 

ros_ntrip_client is flat AI code right?
statusCodeForMessage() looks fragile to me 🤔 .
Beside the endless condition, it parses the status code out of a string message?
That looks fragile to me, without knowing anything about NTRIP and what exact comms is behind the scene 😇 .
So, if it's a useless comment, trash it ;-)

Beside that, already tested it with SAPOS and a couple of AP changes on my critical route. Worked seamless 👍 .
Also did a short RTK2Go test for another guy, because RTK2Go currently doesn't work with our stock NTRIP-Client (some kind of mountpoint error), and that also worked 👍

olliewalsh:
> I'm not lighten enough with NTRIP and ROS to do a valuable review 😇
> 
> ros_ntrip_client is flat AI code right? statusCodeForMessage() looks fragile to me 🤔 . Beside the endless condition, it parses the status code out of a string message? That looks fragile to me, without knowing anything about NTRIP and what exact comms is behind the scene 😇 . So, if it's a useless comment, trash it ;-)
> 
> Beside that, already tested it with SAPOS and a couple of AP changes on my critical route. Worked seamless 👍 . Also did a short RTK2Go test for another guy, because RTK2Go currently doesn't work with our stock NTRIP-Client (some kind of mountpoint error), and that also worked 👍

Yeah, it's horribly messy AI code, and I've made no effort whatsoever to clean it up, but it works.
It's done the status message to status code mapping in a weird way, I'll get that fixed, but it'ss just mapping setStatus(msg) calls to codes so should work fine.

I added my base to rtk2go and centipede.fr to test this, and confirmed that it backs off when the IP gets banned (accidentally, my base had the wrong case for the mountpoint and got my public IP banned).

ClemensElflein:
I'm not sure I'm at the point yet to replace a dependency blindly with "horribly messy AI code" 🤣 

Why do we want to replace the NTRIP client? Because RTK2GO broke? I have provided a fix here for the poeple complaining, but none of them ever bothered to test it: https://github.com/ClemensElflein/open_mower_ros/pull/269. If it's only about that, let's try the simple fix first.

Or are there other larger issues that we need a complete replacement?

ClemensElflein:
Which AI did come up with that? It looks messier than other vibe I've seen

Apehaenger:
Oh, did not know about that RTK2Go fix 👍. Even though I don't use RTK2Go, I'll test today for a guy who get one-on-one support 😪.

The other, for me more relevant issue with the current NTRIP-Client is the problem with WiFi-Roaming. It often doesn't survive AP changes. WebApp, SSH and all other connection survive or reconnect after some seconds of lag, but NTRIP-Client stuck in not getting/sending? rtcm anymore. Only solution for me till now was an `openmower restart` which is okay during a mow session, but annoying if you try to record an area.

olliewalsh:
> Which AI did come up with that? It looks messier than other vibe I've seen

codex/GPT 5.4 (it's cheaper than claude, and this just needs to work)

> The other, for me more relevant issue with the current NTRIP-Client is the problem with WiFi-Roaming. It often doesn't survive AP changes.

^ that

olliewalsh:
Iterated a bit more on this to clean it up, fix a few issues, and add tests. @ClemensElflein could you update the image you built?

rhssk:
> The other, for me more relevant issue with the current NTRIP-Client is the problem with WiFi-Roaming. It often doesn't survive AP changes. WebApp, SSH and all other connection survive or reconnect after some seconds of lag, but NTRIP-Client stuck in not getting/sending? rtcm anymore. Only solution for me till now was an `openmower restart` which is okay during a mow session, but annoying if you try to record an area.

Could be related to this maybe?
https://github.com/ClemensElflein/open_mower_ros/blob/75960119fb72403f652b11812a56605cfaf9824e/src/open_mower/params/openmower_defaults_v2.yaml#L24-L26

These don't seem optimal.

ClemensElflein:
why did we close this?

ClemensElflein:
Ah, because the target `v2` branch is gone, so the PR was auto closed

PR #284, opened 2026-05-01
external_mqtt_topic_prefix 

prigorus:
Problem

Running multiple OpenMower instances with MQTT enabled causes client ID collisions because all instances use the same hardcoded client ID:

ext_xbot_monitoring

This results in:

MQTT disconnect/reconnect loops
delayed or missing updates in Home Assistant

Solution

Use external_mqtt_topic_prefix to generate a unique client ID:

"ext_xbot_monitoring_" + external_mqtt_topic_prefix

Result

Multiple mowers can run simultaneously
Stable MQTT updates (~5s)
No broker conflicts

## Summary by CodeRabbit

* **Chores**
  * External MQTT client identifier is now derived from the topic prefix and constrained to the platform's client-ID length limits to reduce collision risk. Client initialization, callbacks, connection URI and credentials remain unchanged.

### DIFF
--- a/src/lib/xbot_monitoring/src/xbot_monitoring.cpp
+++ b/src/lib/xbot_monitoring/src/xbot_monitoring.cpp
@@ -211,7 +211,7 @@ void setupMqttClient() {
 
         try {
             client_external_ = std::make_shared<mqtt::async_client>(
-                    uri, "ext_xbot_monitoring");
+                    uri, "ext_xbot_monitoring_" + external_mqtt_topic_prefix);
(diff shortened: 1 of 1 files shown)


prigorus (commented):
Edited to generate a unique ID and truncate the length.

prigorus:
@ClemensElflein can this be merged?

ClemensElflein (changes requested):
The hashing is done twice and the external client is created twice, also the URI is gone. I think there might have been an issue during committing this change

ClemensElflein:
stale PR

PR #286, opened 2026-05-10
Allow area recording after manual home

brupje:
If you have pressed home and the mower is docked. After that you cannot run area recording as it is aborting on manual_pause_mowing. This change will detect that manual_pause_mowing is set and clear it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of manual pause flags during area recording initialization. The mower now properly clears pause states when entering area recording mode to ensure correct operation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ClemensElflein/open_mower_ros/pull/286)

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/AreaRecordingBehavior.cpp
@@ -21,6 +21,8 @@ extern actionlib::SimpleActionClient<mbf_msgs::ExePathAction>* mbfClientExePath;
 extern ros::NodeHandle* n;
 extern void registerActions(std::string prefix, const std::vector<xbot_msgs::ActionInfo>& actions);
 
+extern mower_logic::MowerLogicConfig getConfig();
+extern void setConfig(mower_logic::MowerLogicConfig);
 extern void stop();
 
 extern bool setGPS(bool enabled);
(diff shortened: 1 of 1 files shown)


rovo89 (changes requested):
Thanks for your PR, but I object from logical point of view. The flag you're resetting is - among other things - responsible for keeping the mower docked, even when it's fully charged. You can't just reset it.

The question should be why it's even reacting on the flag in this behavior. And I think also on low battery. In area recording mode, the mower is fully controlled by the user, it wouldn't drive home autonomously anyway, so it makes no sense to abort.

So please throw away this change and instead find a clean way to ignore the "drive home" / abort signal.

brupje:
The abort is happening in the safety checks in mower_logic. I guess to protect against automatic mowing after pausing. Kinda makes sense as in most cases you don' t want the mower to just start after the user said it should be docked. 

>  In area recording mode, the mower is fully controlled by the user, it wouldn't drive home autonomously anyway, so it makes no sense to abort.

Sure, but the impact of that change is much bigger. This fix is quite safe to implement and fixes an annoying issue. But feel free to close this PR if you are going to fix it another way.

rovo89:
On the long run, the whole mower_logic node state machine should be refactored... but for now, how about exiting early in checkSafety? It already ignores emergencies in area recording mode. I think it could skip most (if not all) checks there.

ClemensElflein:
Stale PR

PR #287, opened 2026-05-11
Add optional outline_count in map.json for areas

jrv:
This PR adds the option to have outline_count defined in map.json on a per-area basis. It allows the user to set the number of outlines per area.

My specific use case is one area which exists only of narrow paths. Mowing this with infill is very slow, but mowing it as 100% outline works fine. If outline is set to max (255) it will mow as many outlines as possible.

## New Features
* Added per-area outline count configuration capability, allowing customization of coverage planning parameters for individual mowing areas. When not specified, the global configuration value is used as a fallback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ClemensElflein/open_mower_ros/pull/287)

### DIFF
--- a/src/mower_logic/src/mower_logic/behaviors/MowingBehavior.cpp
+++ b/src/mower_logic/src/mower_logic/behaviors/MowingBehavior.cpp
@@ -193,7 +193,8 @@ bool MowingBehavior::create_mowing_plan(int area_index) {
   // calculate coverage
   slic3r_coverage_planner::PlanPath pathSrv;
   pathSrv.request.angle = angle;
-  pathSrv.request.outline_count = config.outline_count;
+  pathSrv.request.outline_count =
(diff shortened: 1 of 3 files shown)


rovo89:
Cool, thanks! I had something like that in mind a while ago, and it was one of the motivations to use JSON for the map. 😊

jrv:
I'm also working on integration of outline_count in openmower-app. It works but it needs some cleaning up before I can send a PR.

rovo89:
> I'm also working on integration of outline_count in openmower-app. It works but it needs some cleaning up before I can send a PR.

I haven't made up my mind yet about how that could look. I've started with a config form, but nothing is set in stone yet. So feel free to send a PR with a suggestion. 😉 You can also just send a screenshot before polishing.

PR #289, opened 2026-05-13
Add opt-in telemetry and crash reporting via Sentry

ClemensElflein:
Introduces anonymous telemetry to help the project ship stable releases faster and surface bugs that only appear in the field.

What's tracked (no GPS or location data ever collected):
- Crashes and C++ exceptions with stack traces (all ROS nodes)
- Mow start/stop, docking, emergency events (from mower_logic)
- Software version and hardware platform per session

Implementation:
- New mower_analytics package builds libmower_analytics_sentry.so, a thin RAII wrapper (SentryGuard) around sentry-native with the crashpad out-of-process backend for reliable crash capture
- SentryGuard linked into mower_logic, monitoring, mower_comms_v1/v2, mower_map_service, and mower_simulation — one line in each main()
- SentryGuard::captureEvent() / addBreadcrumb() callable from anywhere without exposing sentry.h to consumers
- Prints a clear opt-in prompt on startup when OM_TELEMETRY_ENABLE is unset; silent no-op when set to false
- Default DSN pre-configured in the Docker entrypoint; users only need to set OM_TELEMETRY_ENABLE=true in mower_config.sh to participate

## Summary by CodeRabbit

* **New Features**
  * Optional anonymous telemetry and crash reporting (toggle via OM_TELEMETRY_ENABLE).
  * Anonymous events collected: mow start/stop, docking, emergencies, software version, hardware platform; no GPS/location data.
  * Default reporting endpoint can be overridden via OM_SENTRY_DSN.
  * Telemetry integrated into multiple runtime components; processes warn at startup until preference is set.

* **Documentation**
  * Added examples and notes for enabling/disabling telemetry and configuring the reporting endpoint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ClemensElflein/open_mower_ros/pull/289)

### DIFF
--- a/config/mower_config.sh.example
+++ b/config/mower_config.sh.example
@@ -333,6 +333,29 @@ export OM_RAIN_DELAY_MINUTES=30
 # export OM_SNAPSHOT_DURATION=1200
 
 
+################################
+##    Telemetry / Analytics   ##
+################################
+
+# Anonymous telemetry — helps the OpenMower project ship stable releases faster.
+#
+# Enabling this sends crash reports and anonymous usage events (mow start/stop,
+# docking, emergencies, software version, hardware platform) to the OpenMower
+# developers. NO GPS or location data is ever collected.
+#
+# Crash and exception reports help fix bugs quickly. Version stability data lets
+# us see which releases are running reliably in the field and cut stable releases
+# more often.
+#
+# Without a setting the software prints a reminder on every start until you decide.
+#
+# To opt in:
+# export OM_TELEMETRY_ENABLE=true
+#
+# To opt out silently:
+# export OM_TELEMETRY_ENABLE=false
+#
+
 # source the default values for the hardware platform.
 # you only need this line on non-docker installs. in the docker, it will be done automatically.
 # source $(rospack find open_mower)/params/hardware_specific/$OM_MOWER/default_environment.sh
--- a/config/mower_config.sh.example
+++ b/config/mower_config.sh.example
@@ -341,6 +341,29 @@ export OM_RAIN_DELAY_MINUTES=30
 # export OM_SNAPSHOT_DURATION=1200
 
 
+################################
+##    Telemetry / Analytics   ##
+################################
+
+# Anonymous telemetry — helps the OpenMower project ship stable releases faster.
+#
+# Enabling this sends crash reports and anonymous usage events (mow start/stop,
+# docking, emergencies, software version, hardware platform) to the OpenMower
+# developers. NO GPS or location data is ever collected.
+#
+# Crash and exception reports help fix bugs quickly. Version stability data lets
+# us see which releases are running reliably in the field and cut stable releases
+# more often.
+#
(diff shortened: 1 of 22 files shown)


rovo89:
Sounds extremely useful! Apart from crash analytics, I'm especially interested in telemetry, i.e. how many users use which release/git hash/docker image (and which variant). That's mentioned in the summary, haven't checked in the code yet which details exactly are sent.

Do we generate a UUID or so for the mower?

And I assume we'll want to ask the user to opt-in when they first install the OS or use the app, in case they haven't declined yet? Would be good to have a page on openmower.de then to explain what will happen to their data if they agree, so we have one source of truth.

Mower events might conflict with https://github.com/ClemensElflein/open_mower_ros/tree/feat/events-and-position, with the first implementation added in https://github.com/ClemensElflein/open_mower_ros/commit/e0071d7827ccdbcacf09a555a5714071b9573645. How would that compare in implemenation and goal? Do you want to show these events to users or is it more about statistics?

ClemensElflein:
> Sounds extremely useful! Apart from crash analytics, I'm especially interested in telemetry, i.e. how many users use which release/git hash/docker image (and which variant). That's mentioned in the summary, haven't checked in the code yet which details exactly are sent.

Thank you! Yes those are also questions I'd be interested in: Do people even update? Which versions are popular? Is the current edge stable, so that we can tag as release?

In theory I'd also be interested in realistic maps for simulation, so that we can see what the slicer does on actual user maps.



> 
> Do we generate a UUID or so for the mower?

Sentry does this automatically and stores it into the ROS home.

> 
> And I assume we'll want to ask the user to opt-in when they first install the OS or use the app, in case they haven't declined yet? Would be good to have a page on openmower.de then to explain what will happen to their data if they agree, so we have one source of truth.

Yes, the plan is the following:
- No environment variable: Warning in the log file so that people can explicitly set "yes" or "no"
- When `openmower update` (not implemented yet), the tool detects that the analytics envrionment is not yet and asks the user kindly to opt in or out
- Yes, I'll be writing a blog post and an article into the docs on which data is collected and why

> 
> Mower events might conflict with https://github.com/ClemensElflein/open_mower_ros/tree/feat/events-and-position, with the first implementation added in [e0071d7](https://github.com/ClemensElflein/open_mower_ros/commit/e0071d7827ccdbcacf09a555a5714071b9573645). How would that compare in implemenation and goal? Do you want to show these events to users or is it more about statistics?

True it's a bit of a conflict and we should probably merge it and use the same events, also when an emergency is detected we could take a position snapshot of recent history (in local coordinates of course) to see what went wrong.

rovo89:
> also when an emergency is detected we could take a position snapshot of recent history (in local coordinates of course) to see what went wrong

Yes, that was part of the idea for local analytics. I already store the latest position with events, and with the track history, the position history can be reconstructed. Whether we want to analyze that centrally and fix people's gardens... I don't know 😉
Just to be clear, the branch isn't ready for merging yet, needs more testing and we might want to have a closer look at the exact events to be captured.

PR #294, opened 2026-05-21
Make Backward/Forward Recovery behavior look ahead for obstacles, and enable it by default.

jeremysalwen:
1. Enable the recovery behavior by default.
2. Make the recovery behavior check not just the costmap at the current location, but in a line ahead where it's moving to.
3. Manually invoke the recovery behavior in the case where MBF will not recover by itself (specifically when in path following mode).

## Summary by CodeRabbit

* **New Features**
  * Recovery moves now sample and validate points along the full forward/backward path using a configurable obstacle-check distance.
  * Mowing can invoke configured MBF-style recovery behaviors sequentially and stop on the first successful recovery.

* **Bug Fixes**
  * Moves now stop and report failure if an obstacle is detected too close along the recovery path.

* **Chores**
  * Default runtime config enables recovery and adds a backward/forward recovery behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ClemensElflein/open_mower_ros/pull/294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/ftc_local_planner/include/ftc_local_planner/backward_forward_recovery.h
+++ b/src/lib/ftc_local_planner/include/ftc_local_planner/backward_forward_recovery.h
@@ -6,6 +6,7 @@
 #include <tf2_ros/buffer.h>
 #include <base_local_planner/costmap_model.h>
 #include <geometry_msgs/Twist.h>
+#include <geometry_msgs/Pose.h>
 #include <string>
 
 namespace ftc_local_planner
@@ -22,7 +23,7 @@ public:
 
 private:
   bool attemptMove(double distance, bool forward);
-  bool isPositionValid(double x, double y);
+  bool isPathClear(const geometry_msgs::Pose& pose, bool forward);
 
   std::string name_;
   bool initialized_;     
@@ -33,6 +34,7 @@ private:
   double linear_vel_;
   double check_frequency_;
   unsigned char max_cost_threshold_;
+  double obstacle_check_distance_;
   ros::Duration timeout_;
 };
 
--- a/src/lib/ftc_local_planner/src/backward_forward_recovery.cpp
+++ b/src/lib/ftc_local_planner/src/backward_forward_recovery.cpp
@@ -3,16 +3,18 @@
 #include <ros/ros.h>
 #include <tf2/utils.h>
 #include <geometry_msgs/PoseStamped.h>
+#include <cmath>
 
 namespace ftc_local_planner
 {
 
 BackwardForwardRecovery::BackwardForwardRecovery() 
-  : initialized_(false), 
+  : initialized_(false),
     max_distance_(0.5),
-    linear_vel_(0.3), 
-    check_frequency_(10.0), 
+    linear_vel_(0.3),
+    check_frequency_(10.0),
     max_cost_threshold_(costmap_2d::INSCRIBED_INFLATED_OBSTACLE-10),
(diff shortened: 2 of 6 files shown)


Max07B2:
Here are my observations regarding PR-294 (apologies if they sound a bit "vague," but I'm not familiar with the algorithm, and this is what I've noticed):

- I'm testing with Yard Force Classic 500, HWv1, OSv2, open_mower_ros Tag pr-294
- Under normal conditions, everything works as before (so nothing has gotten worse)
- Undocking: (for the sake of completeness) if the mower bumps into something while reversing, no special error handling occurs; I think that’s how it’s always been, and I wouldn’t have expected this PR to change anything here (I’ve never had a problem either...)
- The changed behavior occurs when moving to the next starting position, on the perimeter lines, and on the normal surface => OK
- Behavior when the mower collides
  + It backs up, but then travels almost the same path forward again, so the likelihood of colliding again is very high => I have the feeling that the old PR made a much wider turn here (could that be the case?)
  + Sometimes, for example on the second attempt, it doesn’t back up but tries to turn immediately, which doesn’t work well if the mower is pressed completely against the obstacle
  + If it chooses a slightly different path during retry, it seems to tend to veer to the right, which may be problematic on the outer perimeter lines => if there is an intentional side, I think it should prefer going to the left
  + There seem to be situations where, after a retry, the mower motor doesn't turn on until well past the point where the collision occurred (I would expect it to turn on before it starts moving forward again after reversing) => if this is deterministic at all, I'd consider this a veeery minor issue

I hope this makes sense and, anyway, I consider this feature a good improvement to significantly reduce the situations where the mower gets stuck.

jeremysalwen:
@Max07B2 thanks for the feedback

1. The retry behavior after backing up is the same as before in both PRs (and without the PRs), it will just try to drive to the start of the next point in the path. Note that if we fail we trim points off the path so if there is a curve this *could* mean driving around an obstacle, but if it's straight it's pretty much gonna follow the same path.  It would require marking obstacles to fix that (that honestly might be the best follow up PR).
2. If it's mowing, the sequence should be back up, wait for GPS, attempt to drive to starting point, then, assuming it gets stuck on that, it should back up again, and repeatedly try to drive to the starting point. The only reason I can think it would try to turn after crashing is if it used up its retries, and it's no longer trying to recover, it's just switching to docking mode?
3. There is not an intentional side, it's probably just hardware differences causing the asymmetry?
4. I have another PR from @pete_mi632 that I have which *in general* slows down the mower if the blade stops.  That fixes the problem of blades spinning down for me here (and elsewhere!).

Max07B2:
Thank you very much for the detailed explanation. That explains what I’ve observed, and in that sense, my mower has behaved exactly as intended with your PR. The attempt to navigate around an obstacle would definitely make for a cool follow-up PR (I thought I’d already seen this behavior in the “normal” version, but somehow not always. I’ll take another look at that when I get the chance).

jeremysalwen:
I have fixed the issues found by the AI, and I'm going to test it today on my mower.  One of the issues it found (costmaps not being locked) I think applies in a bunch of places in the codebase, so we should probably fix that!.

jeremysalwen:
It's still running great today!  I think it's ready to merge.

jeremysalwen:
I changed the default parameters to be appropriate for stock openmower, which uses duty cycle instead of m/s for `linear_vel` commands.

PR #305, opened 2026-06-17
fix: publish version to external MQTT broker (fixes #301)

### LINKED ISSUE #301: MQTT: stale retained version message not cleared after upgrade (legacy `version` vs `version/json`)
### Describe the bug

When upgrading OpenMower across an MQTT schema change, old retained messages are not cleared from the broker. The version topic is a concrete example:

- v1.0.2 publishes version on `openmower/version` as `{"version":"v1.0.2"}`
- v1.2.0 publishes version on `openmower/version/json` as `{"version":"v1.2.0"}`

After upgrading from 1.0.2 to 1.2.0, the old `openmower/version` topic keeps its **retained** value (e.g. `v1.0.2`) indefinitely, because nothing clears it. Consumers (e.g. the Home Assistant integration) that read the legacy topic keep showing the outdated version.

### Steps to reproduce

1. Run an older version that publishes to `openmower/version`
2. Upgrade to a version that publishes to `openmower/version/json`
3. Observe that `openmower/version` still holds the old retained value

### Expected behavior

Obsolete retained topics should be cleared (publish empty retained payload) when the schema changes, so stale values don't linger on the broker.

### Suggestion

Add a `schema_version` integer to the `openmower/capabilities` topic, incremented whenever the MQTT schema changes (topics added/removed/renamed, payload format changed). This would let integrations reliably detect schema changes instead of relying on topic-name heuristics, and OpenMower could clean up deprecated retained topics on startup.

### Context

Found while fixing the related issue in the Home Assistant integration (vermut/ha_openmower).

PeterLinuxOSS:
Fixes [#301](https://github.com/ClemensElflein/open_mower_ros/issues/301).

The `version` topic was published only via `try_publish_binary()`, which doesn't publish to the external broker (unlike `try_publish()`). So the external broker (e.g. Home Assistant) kept a stale retained value after upgrades.

This makes `publish_version()` also publish the version as JSON to the external broker, so it always reflects the current value.

*Note: this summary was written by Claude AI.*

## Summary by CodeRabbit

* **New Features**
  * Version information is now published to external MQTT brokers when external monitoring is enabled, providing broader system integration for monitoring purposes.

### DIFF
--- a/src/lib/xbot_monitoring/src/xbot_monitoring.cpp
+++ b/src/lib/xbot_monitoring/src/xbot_monitoring.cpp
@@ -270,7 +270,7 @@ void publish_version() {
     try_publish("version/json", version.dump(), true);
     if(external_mqtt_enable) {
         try {
-            client_external_->publish(external_mqtt_topic_prefix + "version", ersion.dump(), 1, true);
+            client_external_->publish(external_mqtt_topic_prefix + "version", version.dump(), 1, true);
             
         } catch (const mqtt::exception &e) {
             // client disconnected or something, we drop it.
--- a/src/lib/xbot_monitoring/src/xbot_monitoring.cpp
+++ b/src/lib/xbot_monitoring/src/xbot_monitoring.cpp
@@ -268,6 +268,14 @@ void publish_version() {
             {"version", version_string}
     };
     try_publish("version/json", version.dump(), true);
+    if(external_mqtt_enable) {
+        try {
+            client_external_->publish(external_mqtt_topic_prefix + "version", ersion.dump(), 1, true);
+            
+        } catch (const mqtt::exception &e) {
+            // client disconnected or something, we drop it.
+        }
+    }
     auto bson = json::to_bson(version);
     try_publish_binary("version", bson.data(), bson.size(), true);
 }


ClemensElflein:
Thank you for the PR!

rovo89:
@PeterLinuxOSS @ClemensElflein I'm not sure about this fix. Wouldn't `try_publish("version", version.dump(), true);` work the same, without replicating the logic for external MQTT? It looks to me as if we now publish the version to:
- `version/json` as JSON for both internal and external
- `version` as JSON only for external
- `version` as BSON only for internal

And that seems weird.

I "broke" this in https://github.com/ClemensElflein/open_mower_ros/commit/a57cbd06541a49c127b8c5c98ec62bfbf219dd47, as it was inconsistent with other topics, which mostly use the `/json` suffix for `try_publish()` and `/bson` for `try_publish_binary()`. If I started a project today, I would consider JSON as standard and if really necessary only have `/bson` use a suffix (or skip it all all). But it's probably not worth changing all topics... I thought `version` is small enough to fix it.

rovo89:
Ah, and it seems that we published both JSON and BSON to the same `version` topic before, for both internal and external?!? Maybe I had issues with the app receiving the binary payload and that was the motivation to split it. I'm not sure anymore, it was months ago.

PeterLinuxOSS:
> Ah, and it seems that we published both JSON and BSON to the same `version` topic before, for both internal and external?!? Maybe I had issues with the app receiving the binary payload and that was the motivation to split it. I'm not sure anymore, it was months ago.

It was the simplest solution to this problem because I was not sure why, in https://github.com/ClemensElflein/open_mower_ros/commit/a57cbd06541a49c127b8c5c98ec62bfbf219dd47, it was removed and placed in json. As a result, `version` was reporting the old version instead of the actual one because it was moved to `version/json`.

ClemensElflein:
Thank you @rovo89 for double checking, sorry if I merged prematurely. I didn't have https://github.com/ClemensElflein/open_mower_ros/commit/a57cbd06541a49c127b8c5c98ec62bfbf219dd47 in mind.

Shall I revert and reopen?

ClemensElflein:
Yes, bson and json going to the same topic is not a good idea.

rovo89:
> Shall I revert and reopen?

Not yet - we should just discuss what the final state should be. My pick would be to keep `version/json` for consistency and clear the retained value for `version` once during startup. Then at least clients looking at the matter won't get outdated info. If it helps, we can add a capability for the version... but not sure how that would make it better. Mich easier to subscribe to both and use the older one as fallback.

Schema version as capability isn't how it's intended. Capabilities should reflect features, not overall app version. That way, for example during development we can check exactly for that feature (or important milestones of it), and it's trivial to merge if it even conflicts at all.

PeterLinuxOSS:
> > Shall I revert and reopen?
> 
> Not yet - we should just discuss what the final state should be. My pick would be to keep `version/json` for consistency and clear the retained value for `version` once during startup. Then at least clients looking at the matter won't get outdated info. If it helps, we can add a capability for the version... but not sure how that would make it better. Mich easier to subscribe to both and use the older one as fallback.
> 
> Schema version as capability isn't how it's intended. Capabilities should reflect features, not overall app version. That way, for example during development we can check exactly for that feature (or important milestones of it), and it's trivial to merge if it even conflicts at all.

Yes, it’s a good idea to clear it at startup. Additionally the code https://github.com/vermut/ha_openmower/pull/12 is structured so that `version` will become obsolete and be replaced by `version/json`. But I wasn't sure initially if the failure to publish the version to external MQTT was intentional or just a bug...

PR #309, opened 2026-06-18
Optimize fill path traversal as for obstacles

bodop:
We do the same optimization here as for obstacles: Always continue with the next fill path segment nearest to the current mower position. Optionally reverting the direction of the segment.

## Release Notes

* **Performance & Optimization**
  * Coverage path planning now improves route efficiency by using a “last visited” position to greedily reorder fill-line traversal and automatically reverse line direction when beneficial.


ClemensElflein:
Can you share your map where you are running into issues? The slic3r code should already output the lines in a sensible order AFAIK

bodop:
[map.json](https://github.com/user-attachments/files/29097764/map.json)
Due to the "U" shaped perimeter, dependent on the mowing angle, the mower often hops from one limb to the other.

PR #310, opened 2026-06-18
Nonlinear angular control

jeremysalwen:
A patch from Pete_MI632 to improve the PID control of the mower.

Basically, we reduce the strength of the angular PID terms when the lateral distance is large.  So effectively you can have good angular control when on the track, but you can still have responsive lateral control when you have wandered away from the track.

The formula is pretty straightforward, it just linearly scales down the strength of the angular PID terms from 100% at 0.0m of lateral error, to 10% at 9cm of lateral error, and beyond 9cm it remains at 10% strength. The lateral_priority_distance configures this 9cm distance*.  You can also disable this features entirely by setting lateral_priority_distance to 0.

I have been running with this setting for the past year or so, and it works well.  Anyone who is running my Mowgli-docker setup, and of course @Pete_MI632  has tested this patch as well.

## Summary by CodeRabbit

* **New Features**
  * Added `lateral_priority_distance` to automatically reduce angular PID correction based on lateral error, scaling down to zero at the configured threshold (can be disabled with `0`).
  * Added `lateral_priority_min_gain` to clamp the angular PID gain so it won’t drop below a minimum when the scaling is active.
* **Configuration Updates**
  * Extended default planner configuration for real, docking, and simulation modes with `lateral_priority_distance: 0.1` and `lateral_priority_min_gain: 0.1`.


ClemensElflein:
Do you think this one is ready to go?

jeremysalwen:
I think it's ready to go now,  multiple people have tested it.

ClemensElflein:
Nice 👍 thank you for reporting back. Is there actually noticable control improvement or is it more of a theoretical fix?

jeremysalwen:
Pete has reported a control improvement when he originally implemented it,
technically he he using a different (but basically equivalent) patch.

Other than that I don't think anyone has done a real apples to apples
comparison.

ClemensElflein:
In any case, it does make sense. In cases like this telemetry would be nice then we could actually see some before and after traces

PR #25, opened 2023-05-01
added custom angle to mower_config.txt 

NDaub:
added custom angle to mower_config.txt to manually set the mowing angle

PR #324, opened 2026-07-06
mower_comms_v2: drive the mow motor with a signed speed

rpeltola:
Sends a single signed `Mower Speed` to the LL board (enable + direction mapped to +1 / -1 / 0), pairing with the firmware PR after the `MowerDirection` register was dropped (definitions #25, merged).

- Direction now rides on the sign of `mower_motor_rpm`; drops the separate `Status.mow_direction` field and its handler (monitoring takes `abs()`).
- Between-run direction randomization is now opt-in (`randomize_mow_motor_direction`, default off).
- Bumps the `services` submodule to the merged definitions (`c074039`).

### DIFF
--- a/src/mower_comms_v2/src/MowerServiceInterface.cpp
+++ b/src/mower_comms_v2/src/MowerServiceInterface.cpp
@@ -5,12 +5,14 @@
 #include "MowerServiceInterface.h"
 
 void MowerServiceInterface::Tick() {
-  SendMowerEnabled(status_msg_.mow_enabled);
+  SendMowerSpeed(commanded_speed_);
 }
 
-void MowerServiceInterface::SetMowerEnabled(bool enabled) {
-  SendMowerEnabled(enabled);
-  status_msg_.mow_enabled = enabled;
+void MowerServiceInterface::SetMowerSpeed(float speed) {
+  // Signed speed/duty in [-1, 1]: sign = direction, 0 = off.
+  commanded_speed_ = speed;
+  SendMowerSpeed(speed);
+  status_msg_.mow_enabled = speed != 0.0f;
   status_publisher_.publish(status_msg_);
 }
 
@@ -44,7 +46,9 @@ void MowerServiceInterface::OnMowerMotorRPMChanged(const float& new_value) {
 
 void MowerServiceInterface::OnServiceConnected(uint16_t service_id) {
   status_msg_ = {};
-  SendMowerEnabled(false);
+  // Clear the cached speed so a stale value isn't replayed after reconnect.
+  commanded_speed_ = 0.0f;
+  SendMowerSpeed(commanded_speed_);
 }
 
 void MowerServiceInterface::OnTransactionStart(uint64_t timestamp) {
--- a/src/mower_logic/src/monitoring/monitoring.cpp
+++ b/src/mower_logic/src/monitoring/monitoring.cpp
@@ -18,6 +18,8 @@
 #include <mower_msgs/Power.h>
 #include <xbot_msgs/SensorDataString.h>
 
+#include <cmath>
+
 #include "mower_logic/MowerLogicConfig.h"
 #include "mower_logic/PowerConfig.h"
 #include "mower_logic/utils.h"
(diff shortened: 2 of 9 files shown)


src/mower_comms_v2/src/MowerServiceInterface.cpp
ClemensElflein, line 1: Instead of sending two messages to the firmware, please add the message for the MowerEnabled input instead, so that it takes a speed instead. Let's use float or double or something and send a value between -1.0 and 1.0.

Since we need to change the mower definitions anyways, this will require firmware update anyways, so we should keep it one single message.

~~Alternatively this also can be an RPC call (xbot_framework now supports RPC calls since a few weeks)~~

ClemensElflein (changes requested):
Thank you for the PR! Good idea in general, but please adapt the ROS <--> Firmware message as described in the comment.

src/mower_comms_v2/src/MowerServiceInterface.cpp
rovo89: I'm not sure if RPC will would be right for this. Isn't a call more like a one-off thing, whereas mower speed is a (temporary) setting?

ClemensElflein: You are right, let's keep it as normal input

rpeltola: Changed to float input. I kept the MowerDirection enum for telemetry (on the firmware side). Is it okay?

rovo89: Do you fear that the mower will sometimes not react or delays the command? What's the reason to basically read back the input value?

rpeltola: I though it's nice to have actual firmware report the direction for telemetry uses. Before changing the input to float I had block in the firmware that did not allow switching the direction when the motor was running. I removed the block to simplify it. Not sure if it's up the firmware to block such input or not.

ClemensElflein:
Should we make this configurable (is it already)?

Otherwise LGTM, let's finalize the firmware and then it's good to go

src/mower_logic/src/monitoring/monitoring.cpp
```diff
@@ -85,7 +87,7 @@ std::map<std::string, SensorConfig> sensor_configs{
...
   {"om_mow_motor_temp", {"Mow Motor Temp", "deg.C", xbot_msgs::SensorInfo::VALUE_DESCRIPTION_TEMPERATURE, xbot_msgs::SensorInfo::TYPE_DOUBLE, [](StatusPtr msg) { return msg->mower_motor_temperature; }, &set_limits_mow_motor_temp, "mower_xesc", [](){ return paramNh->param("mower_xesc/has_motor_temp", true); }}},
   {"om_mow_motor_current", {"Mow Motor Current", "A", xbot_msgs::SensorInfo::VALUE_DESCRIPTION_CURRENT, xbot_msgs::SensorInfo::TYPE_DOUBLE, [](StatusPtr msg) { return msg->mower_esc_current; }, &set_limits_mow_motor_current, "mower_xesc"}},
-  {"om_mow_motor_rpm", {"Mow Motor Revolutions", "rpm", xbot_msgs::SensorInfo::VALUE_DESCRIPTION_RPM, xbot_msgs::SensorInfo::TYPE_DOUBLE, [](StatusPtr msg) { return msg->mower_motor_rpm; }, &set_limits_mow_motor_rpm, "mower_xesc"}},
+  {"om_mow_motor_rpm", {"Mow Motor Revolutions", "rpm", xbot_msgs::SensorInfo::VALUE_DESCRIPTION_RPM, xbot_msgs::SensorInfo::TYPE_DOUBLE, [](StatusPtr msg) { return std::fabs(msg->mower_motor_rpm); }, &set_limits_mow_motor_rpm, "mower_xesc"}},
```
Apehaenger: What's against having monitoring also report the direction by signed RPMs?
It was a good indicator in past for badly configured ESCs.

Apehaenger:
Quite thanks for this PR and your effort to make the devs happy with their comments :innocent: 

Due to "rebase to current main" requirement and missing permissions for that in your repo, I pushed your PR to #334

PR #328, opened 2026-07-13
Log emergency reasons

ClemensElflein:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved emergency status reporting by logging when an emergency is cleared or when its reason changes.
  * Emergency state messages continue to be published unchanged.


rovo89:
@ClemensElflein This is mainly for debugging, right? In the future, I hope we can adjust mower_logic in a way that we can keep track of different emergency reasons and trigger events for them. Had a draft for that already, but realized that it was hard to integrate into the current SSM (spaghetti state machine).

ClemensElflein:
Yes one User has random High Level emergencies and it would be interesting to see when exactly it happens and what the previous logs were

PR #329, opened 2026-07-13
Re-calibrate gyro-calibration while the robot is standing still.

bodop:
Calibration of the gyroscope gets lots over time, e.g. due to thermal effects. As a consequence the robot seems to rotate slowly, while it tries to stand still (e.g. in the docking station).

This patch assumes, that the robot tries to stand still, when there are no wheel ticks. This time is used to re-calibrate the gyroscope. The robot's orientation is not lost during charging, which is important for perimeter docking, as the robot's pose cannot be reset to the recorded pose then.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gyroscope calibration while the vehicle is stationary.
  * Added detection of external movement during calibration to prevent inaccurate readings.
  * Improved vehicle speed and motion handling using velocity input.
  * Removed reliance on wheel-tick data for motion updates.
  * Enhanced stationary-state detection and recalibration accuracy.

### DIFF
--- a/src/lib/xbot_positioning/src/xbot_positioning.cpp
+++ b/src/lib/xbot_positioning/src/xbot_positioning.cpp
@@ -41,12 +41,25 @@ xbot_msgs::WheelTick last_ticks;
 bool has_gps;
 xbot_msgs::AbsolutePose last_gps;
 
-// True, if last_imu is valid and gyro_offset is valid
+// True, if last_imu is valid
 bool has_gyro;
 sensor_msgs::Imu last_imu;
 ros::Time gyro_calibration_start;
-double gyro_offset;
-int gyro_offset_samples;
+static double gyro_offset=0;
+
+// Needed for offset calibration
+static double gyro_offset_s=0;
+static double gyro_offset_s2=0;
+static int gyro_offset_samples=0;
+
+// We consider to be at rest, if no activity has been detected for this number of seconds.
+#define FROZEN_DELAY 5
+// Samples needed for averaging
+#define GYRO_OFFSET_SAMPLES 500
+// How long did we not move
+static double frozen_duration=FROZEN_DELAY;
+// Maximum velocity, that we accept as standing still
+static double frozen_angular_max_velocity=1000;
 
 // Current speed calculated by wheel ticks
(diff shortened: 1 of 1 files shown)


ClemensElflein:
wouldn't it make more sense to recalibrate while we know it will be stationary? I.e. while in idle? maybe directly before starting the mow?

bodop:
I think a positive side effect of the current solution is, that it works even when the robot stands still in a narrow passage with lost RTK fix. It still keeps the correct orientation. Otherwise the robot would virtually rotate and start moving in the wrong direction after RTK is back again.

ClemensElflein:
But if calibration is done right before RTK comes back, then it's an issue, right? I mean we can also do it during PAUSE state, before transitioning to mowing.

In reality this is only an issue, because the calibration currently is only done once per startup, so when it's in PAUSE and resumes, it potentially has a calibration from a few days ago.

I think calibrating before starting each mow cycle is more than enough

bodop:
For me the calibration itself isn't an issue at all. While the robot is in motion, the kalman filter will always correct the orientation. For me the problem is, that the robot seems to rotate (very slowly), while it is stopped. Why not suppressing this rotation completely and use the data for simultaneous correction of the calibration. Of course also the PAUSED state could be used for this, but an extra calibration phase after that would introduce a needless delay.

ClemensElflein:
- Sounds exactly like offset calibration is the issue you are facing though. The gyro drift is calibrated once and after long time of use, it's not accurate anymore which causes slow rotation on stand-still when PAUSED.
- I understand that the current "once per container lifetime"-calibration leads to drift after long use.
- Calibrating each time before starting the mow is then probably more than enough and doesn't introduce noticeable delay. On-the-fly calibration feels hacky and happens unnecessarily often.
- We can add a "if no wheel ticks, then gate slow rotation values below a certain threshold" in addition. I just don't like the continuous recalibration

If I'm not understanding the issue correctly, please explain where my understanding is wrong

bodop:
Your understanding is correct. The only question is, if there is a need for an additional option, or if there exists only one reasonable setting for it. The remaining things are personal preferences.

PR #331, opened 2026-07-16
Enhance Docker build process to support software versioning via build arguments

PeterLinuxOSS:
### Problem
After updating to `master`, the reported version (app + MQTT `openmower/version` /
`openmower/version/json`) shows `v0.0.0` instead of the real version.

### Root cause
#321 added a `.dockerignore` entry excluding `.git/` from the Docker build context, but the
Dockerfiles still compute the version via `git describe` at build time. Without `.git`, that
call fails and every image silently falls back to `v0.0.0`.

### Fix
- Added an `OM_SOFTWARE_VERSION` build-arg to both Dockerfiles, used when provided, falling back
  to `git describe` (local builds), then `v0.0.0`.
- CI now computes the version from git and passes it as a build-arg to all image variants.
- `.dockerignore` left untouched — `.git/` should stay out of the build context for caching.

---
*Note: summary was generated with a Claude AI.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version identification for container images built in continuous integration.
  * Ensured version information remains available when source-control metadata is unavailable.
  * Added consistent version handling across standard and legacy container image builds.
  * Improved the reliability and traceability of published images by preserving accurate version details throughout the build process.
  * Container images now retain a dependable version identifier across supported build environments, making releases easier to identify and verify.


ClemensElflein:
Superseded by #350 — rebased onto current main (original branch had unrelated/merge commits mixed in).

PR #344, opened 2026-08-23
Fix IPv6 disabled for remote

amylizzle:
Fixes the a crash in `xbox_remote`  when IPv6 is removed from the kernel, rather than disabled (thanks ecovacs!)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket server startup reliability.
  * Added a fallback to IPv4 when the default connection attempt fails.
  * Added clearer handling for startup failures.


amylizzle:
testing for the exact error won't work because websocketpp wraps its errors in a generic "underlying transport error". Best just to try again and fail gracefully if it still doesn't work.

PR #350, opened 2026-09-07
Enhance Docker build process to support software versioning via build arguments

ClemensElflein:
Rebased/cleaned version of #331 onto current `main` (original branch had unrelated merge commits and reverted commits mixed in — this PR contains only the actual Docker versioning change).

### Problem
After updating to `main`, the reported version (app + MQTT `openmower/version` / `openmower/version/json`) shows `v0.0.0` instead of the real version.

### Root cause
#321 added a `.dockerignore` entry excluding `.git/` from the Docker build context, but the Dockerfiles still compute the version via `git describe` at build time. Without `.git`, that call fails and every image silently falls back to `v0.0.0`.

### Fix
- Added an `OM_SOFTWARE_VERSION` build-arg to both Dockerfiles, used when provided, falling back to `git describe` (local builds), then `v0.0.0`.
- CI now computes the version from git and passes it as a build-arg to all image variants.
- `.dockerignore` left untouched — `.git/` should stay out of the build context for caching.

Closes #331

## Summary by CodeRabbit

* **Improvements**
  * Docker images now include consistent software version metadata.
  * Version information uses available release identifiers with reliable fallback values when detailed history is unavailable.
  * Version metadata is applied consistently across current and legacy image builds.
  * Image versions are now more predictable and easier to identify across releases and build environments.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -83,6 +83,17 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: 'recursive'
+          # Full history + tags so `git describe` can resolve the real version below.
+          fetch-depth: 0
+
+      # Compute the software version here (git is available in the runner, but the
+      # Docker build context excludes .git/, so the container can't compute it itself).
+      - name: Compute version
+        id: version
+        run: |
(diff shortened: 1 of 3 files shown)


ClemensElflein:
@coderabbitai review

ClemensElflein:
@coderabbitai full review

